PR #22, opened 2020-02-12
cmake: Silence warnings with clang.

orbea:
This adds a `check_cxx_flag` function to ` CMakeModules/CheckCXXFlag.cmake` designed to make it easier to check compiler flags across the build system and implemented for the existing `-Wall` check, `-Wno-class-memaccess` and `-Wno-invalid-offsetof`. There may be other places this can be used, but I think this is enough for now.

I also removed the existing msg for enabling `-Wall` because cmake seems adequate to informing the result already.
```
-- Performing Test COMPILER_SUPPORTS_WALL
-- Performing Test COMPILER_SUPPORTS_WALL - Success
-- Performing Test COMPILER_SUPPORTS_MEMACCESS
-- Performing Test COMPILER_SUPPORTS_MEMACCESS - Failed
-- Performing Test COMPILER_SUPPORTS_OFFSETOF
-- Performing Test COMPILER_SUPPORTS_OFFSETOF - Success
```

Also note its important to use `-Werror` for these checks, because the compiler may only emit an warning and not an error if they are not supported.

The goal behind this is to silence these numerous compiler warnings with `clang-9.0.1`.
```
warning: unknown warning option '-Wno-class-memaccess'; did you mean '-Wno-class-varargs'? [-Wunknown-warning-option]
1 warning generated.
```


orbea:
err, this is not quite right yet....

orbea:
For some reason the build is not getting `-Wno-class-memaccess'` or `-Wno-invalid-offsetof` even when the tests pass when using `-GNinja`, but they are used correctly with cmake's make generator. I don't understand why yet.

orbea:
The `set` command was missing `PARENT_SCOPE`. It should be fine now.

stenzek:
Sounds good, thanks!

PR #285, opened 2020-04-14
AppImage: Add desktop integration

ggdrt:
PR adds desktop integration to AppImages. Looks like this on Xfce 4:

Other changes include increasing the number of available icon sizes to better support different desktop environments and splitting the AppImage generation out to its own script so users can build locally if they wish.


appimage/generate-appimages.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+function get_tools_and_plugins() {
+  wget https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage
```
stenzek, line 4: If a user wants to build locally like you mentioned in the description and updates their repo, would it be worth checking for the existence of this file? Otherwise I believe `wget` will download a second copy.

.gitignore
```diff
@@ -5,6 +5,7 @@
 /bin/
 /Build/
 /build/
+/build-*/
```
stenzek, line 8: Might want to add the appimage deploy scripts here as well (or create a `.gitignore` in the appimage directory for them)

ggdrt:
Should be more robust now. Script now uses the user-specified build directory as the download directory for linuxdeploy and any AppDir work it does. The catch is that any AppImages are going to be generated in the directory the generate script is called from. It doesn't look like linuxdeploy has any option to specify where to output it to, so we just manually copy the generated AppImages into the build directory. 

Linuxdeploy also doesn't strip the duckstation executables or any shared libraries, but that's probably not something to worry about.

stenzek:
Fantastic, thank you!

> Linuxdeploy also doesn't strip the duckstation executables or any shared libraries, but that's probably not something to worry about.

We could probably do this as part of the cmake build process if it becomes an issue. But regardless, I think a ~30mb download size is fine, especially since it includes Qt.

PR #474, opened 2020-05-23
AppImage: Add zsync support

ggrtk:
The embedded update information in the AppImages is hardcoded to use GitHub rolling releases. The URLs used [shouldn't be changed](https://github.com/AppImage/AppImageSpec/blob/da2633c082fcd83a60aa3a28381e5c0b49ef7bed/draft.md#update-information) unless absolutely necessary I believe, e.g. if migrating to an organization repository.

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -67,11 +67,23 @@ jobs:
         name: "linux-x64-appimage-sdl"
         path: "build/duckstation-sdl-x64.AppImage"
 
+    - name: Upload SDL AppImage zsync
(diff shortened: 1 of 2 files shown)


ggrtk:
Urgh, going to remove the last commit. Not sure why I thought to package the zsync files in the archive with the AppImage. It's pointless since the embedded update info in the AppImages isn't going to look for the zsync files from AppVeyor anyways and having those files doesn't do us any good.

stenzek:
Whoops, seems I need to adjust the Actions script to not try to create a release for pull requests. But that's my error :)

LGTM, thank you for updating the script.

PR #506, opened 2020-05-28
Update input profiles and input profile handling

ggrtk:
Sort of a multi-facted PR for a couple of small things.

Of note, the updated DualShock 4 input profile also works for DualShock 3 on Linux, although I wasn't able to get it to work with the DualShock 3 on Windows. There's probably some driver-related shenanigans going on there. Rumble doesn't work for either gamepad on Windows, but I think this is expected for non-XInput devices on Windows going through SDL.


stenzek:
Sounds good. Thanks!

I figured at some point I'll have to write xinput/directinput backends, I was just hoping to hold off for as long as possible :)

PR #516, opened 2020-05-31
Controller: Add NeGcon support

ggrtk:


stenzek:
Looks good, thanks! Will fix the MSVC build afterwards.

PR #520, opened 2020-06-02
Update GitHub Actions workflow

ggrtk:
This change should stop PRs from running and failing the create-release job. Checkout action v2 should also be faster than v1 due to its default single commit fetching behavior and would probably be helpful as the repository grows.

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -88,6 +88,7 @@ jobs:
   create-release:
     needs: [windows-build, linux-build]
     runs-on: "ubuntu-latest"
+    if: github.ref == 'refs/heads/master'
     steps:
       - name: Download Windows x64 Artifact
         uses: actions/download-artifact@v1


stenzek:
My only concern here would be that the scm version wouldn't get tagged correctly if it's doing a shallow clone (since it won't get the 0.1 tag and history that it's relative to).

ggrtk:
Hmm right. I'll check if there's a better solution that doesn't break that feature.

stenzek:
Given it's hosted on the same infrastructure I don't think it should be too expensive to clone. But in theory we could just fetch until the last not-"latest" tag and that should give the correct revision number (since it's relative to that, currently "0.1").

ggrtk:
Sorry for jumping the gun earlier -- went ahead and removed that commit, looks like there's still some discussion over in the actions/checkout repo for this behavior. Seems best to just leave as-is for now instead of trying to update and work around it.

PR #538, opened 2020-06-07
CMake: Always run gen_scmversion script on rebuilds

ggrtk:
Another small housekeeping fix for an issue mentioned in Discord. I think it's also possible to use a CMake script for generating the scmversion file instead of using separate batch and shell scripts, but that's probably unnecessary right now.


stenzek:
I'd be fine with doing it purely in cmake. Thanks for the workaround.

PR #549, opened 2020-06-14
Qt: Simplify Rebind All behavior

ggrtk:
The current behavior for Rebind All is to unconditionally rebind all buttons regardless of what the user selects in the question message box, with the selection only determining whether or not bindings are cleared before starting the full rebinding process. If Rebind All was pressed unintentionally, it seems that the user has to wait through the timeout process to prevent bindings from being changed.

To make this less prone to user error and more straightforward, Rebind All should first ask for user confirmation (like Clear All does), and if the user confirms, clear all bindings and run the rebinding.


stenzek:
Sounds reasonable.

PR #561, opened 2020-06-23
rectangle: Correct assignment of right within SetExtents

lioncash:
Previously left was being assigned twice in a row.

Thankfully, this function wasn't used anywhere yet, so no code was impacted by this =)


stenzek:
Yikes, I probably should feed it through a static analyzer, it'd probably catch some of these..

PR #562, opened 2020-06-23
spu: Make use of logical AND within IsPitchModulationEnabled()

lioncash:
It seems awfully suspect to use a bitwise AND here.


stenzek:
Was probably deliberate as both sides are bool, but I'd expect any compiler would do this on its own.

PR #564, opened 2020-06-23
file_system: Prevent resource leaks in DeleteDirectory()

lioncash:
We need to remember to close the search handle in the error cases.


stenzek:
This is code from the early 2010s - probably should be replaced with a unique_ptr/deleter... one day I'll go through and modernize it all.

PR #566, opened 2020-06-23
ControllerInterface: Customizable axis scaling and deadzone size

ggrtk:
Feature/workaround to accommodate for differences between the logical range of reporting on the analog sticks of modern controllers and original PS1 analog controllers. Ballpark for a good scaling value is somewhere around 1.30~1.40 but every controller is a bit different so it'll be mostly up to user preference.

### DIFF
--- a/src/frontend-common/common_host_interface.cpp
+++ b/src/frontend-common/common_host_interface.cpp
@@ -1033,6 +1033,9 @@ void CommonHostInterface::UpdateControllerInputMap(SettingsInterface& si)
 
     const float axis_scale = si.GetFloatValue(category, "AxisScale", 1.00f);
     m_controller_interface->SetControllerAxisScale(controller_index, axis_scale);
+
+    const float deadzone_size = si.GetFloatValue(category, "Deadzone", 0.25f);
+    m_controller_interface->SetControllerDeadzone(controller_index, deadzone_size);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/frontend-common/common_host_interface.cpp
```diff
@@ -1030,6 +1030,12 @@ void CommonHostInterface::UpdateControllerInputMap(SettingsInterface& si)
       for (const std::string& binding : bindings)
         AddRumbleToInputMap(binding, controller_index, num_motors);
     }
+
+    const float axis_scale = si.GetFloatValue(category, TinyString::FromFormat("AxisScale"), 1.00f);
```
stenzek, line 1034: This can just be `si.GetFloatValue(category, "AxisScale", 1.00f)`.

src/frontend-common/common_host_interface.cpp
```diff
@@ -1030,6 +1030,12 @@ void CommonHostInterface::UpdateControllerInputMap(SettingsInterface& si)
...
+    m_controller_interface->SetControllerAxisScale(controller_index, axis_scale);
+
+    const float deadzone_size = si.GetFloatValue(category, TinyString::FromFormat("Deadzone"), 0.25f);
```
stenzek, line 1037: Same.

ggrtk:
Fixed, thanks!

PR #567, opened 2020-06-23
HostInterface: Save default setting to AspectRatio, not PixelAspectRatio

ggrtk:
Minor fix. Also went ahead and hooked up the aspect ratio setting in LibretroHostInterface since why not I guess.


src/duckstation-libretro/libretro_host_interface.cpp
```diff
@@ -81,7 +80,7 @@ void LibretroHostInterface::retro_get_system_av_info(struct retro_system_av_info
 
   std::memset(info, 0, sizeof(*info));
 
-  info->geometry.aspect_ratio = 4.0f / 3.0f;
+  info->geometry.aspect_ratio = Settings::GetDisplayAspectRatioValue(m_settings.display_aspect_ratio);
```
stenzek, line 83: This actually gets overridden by the GPU later on, but it'll end up with the same value (https://github.com/stenzek/duckstation/blob/pending/src/duckstation-libretro/opengl_host_display.cpp#L338)

ggrtk: Yeah, unfortunately though frontends typically use the first call of retro_get_system_av_info to set the windowed dimensions, requiring the user to manually do a fullscreen/windowed mode toggle for things to show up at their preferred aspect ratio. This change allows us to report the user-defined aspect ratio on the first call, to avoid users having to manually work around this frontend behavior.

PR #588, opened 2020-07-05
GPU: CRTC fixes

ggrtk:
Adjusted alignment formula doesn't cause any regression in 0ef0c36, and shouldn't cause any problems in well-conditioned cases.


stenzek:
Fantastic! Thank you for debugging this.

PR #617, opened 2020-07-17
Qt: Replace QtSettingsInterface with INISettingsInterface

ggrtk:
One thing I'm concerned about is that users will have their settings deleted without much warning, since a pop-up window doesn't seem to get created when a settings version mismatch is found, like it does when there is a save state version mismatch.


stenzek:
Fantastic, thanks for doing this. Will have a thorough look when I get a chance.

PR #622, opened 2020-07-18
Qt/GameListSettings: Use native path separators

ggrtk:
`QFileDialog::getExistingDirectory()` returns paths with forward slash by default. This path string eventually gets passed to `FileSystem::FindFiles()` which constructs the rest of the path string using native path separators. This means that for example if on Windows we had added `C:/path/to/game` non-recursively and `C:/path/to` recursively through Qt, we would find `C:/path/to/game\x.cue` and `C:/path/to\game\x.cue`, resulting in the same game appearing twice on the game list.


stenzek:
Nice catch!

PR #625, opened 2020-07-18
Disable Interlacing default as "Unchecked"

mirrornoir:
As noted by Kurayami6, Disable Interlacing should show "Unchecked" as the default option for this enhancement.


stenzek:
Thanks! :)

Issue #638, opened 2020-07-21
Crashes when try to use a controller button as a hotkey


Shideravan:
### Version
DuckStation 0.1-1304-gc8b6712

### What happens?
The emulation crashes when trying to use a controller button on hotkey.

### What *should* happen?
Not to crash and permit to use controller button as a hotkey.

### What hardware, operating system, and Duckstation version?  On desktop, GPU matters for graphical issues.
I7-4930, 16GB RAM, GTX 1070
Windows 10 x64
Using PS3 controller with SCPToolkit to simulate a XInput controller.
In the controller bindings the issue doesn't happen, only in the hotkey.

ggrtk:
This should be fixed as of 855fc9a.

stenzek:
Please let me know if this is still happening.

Shideravan:
Now it's fine, thank you 😊

PR #641, opened 2020-07-22
CommonHostInterface: Prevent some potential null pointer dereferences

ggrtk:
This should fix issue #638. 

Might also need to block controller polling for a short period of time or something to that effect to prevent the hotkey functionality from being engaged instantly after being registered, but that's a separate issue from the crash.


stenzek:
Ah yes, this was also an issue in SDL if you hit tab while there was no game active. Kept forgetting to fix it, thanks.

PR #680, opened 2020-08-01
(libretro/Vulkan) Fix glslang shader-related crashes

### LINKED ISSUE #672: [libretro - Linux - Vulkan - Intel iGPU] Crash on successive load content
When running the libretro core under Linux using the Vulkan backend with Intel integrated graphics, the first 'load content' operation works correctly - but attempting to load content a second time within the same session results in a crash, with the following assertion:

```
Assertion failed: 'Context should have been destroyed by now' in function ~VulkanHostDisplay (/home/buildbot/buildbot/linux_x86_64/libretro-duckstation/src/frontend-common/vulkan_host_display.cpp:95)
Aborting application.
Aborted (core dumped) 
```

I get the same result on two machines:

- OpenSUSE Leap 15.2 + Intel HD Graphics 5500
- OpenSUSE Tumbleweed + Intel UHD Graphics 605

However, the issue does not manifest under Linux when using Nvidia graphics (with the official binary blob driver) - content can be launched any number of times, and no errors occur. So it may be the case that this is an Intel driver bug...

jdgleaver:
As reported in #672, the libretro core crashes when loading more than one item of content within a single session when running under Linux with an Intel integrated GPU.

There were actually 2 issues here:

1) `glslang::FinalizeProcess()` is attached via `std::atexit()`. This does not work as intended when running duckstation as a dynamic shared object library... (it only gets called when RetroArch quits, not when the core is unloaded)

2) With Linux + Intel iGPU, I was getting hit by the same 'mysterious' corruption issue that affected RetroArch itself some time ago, and which was fixed in this commit: https://github.com/libretro/RetroArch/commit/4437cd1eac144f1a3b5acda54c128213e59ec0e1

This PR ensures that `glslang::FinalizeProcess()` is called when unloading the core (and not on RetroArch exit), and backports the changes in https://github.com/libretro/RetroArch/commit/4437cd1eac144f1a3b5acda54c128213e59ec0e1. This fixes all crashing issues on my systems.

This closes #672

@stenzek I am by no means an expert in this particular area, so please review this carefully before merging! :)

### DIFF
--- a/dep/glslang/OGLCompilersDLL/InitializeDll.cpp
+++ b/dep/glslang/OGLCompilersDLL/InitializeDll.cpp
@@ -158,6 +158,7 @@ bool DetachProcess()
 
     OS_FreeTLSIndex(ThreadInitializeIndex);
     ThreadInitializeIndex = OS_INVALID_TLS_INDEX;
+    DeinitializePoolIndex();
 
     return success;
 }
--- a/dep/glslang/glslang/Include/InitializeGlobals.h
+++ b/dep/glslang/glslang/Include/InitializeGlobals.h
@@ -38,6 +38,7 @@
 namespace glslang {
 
 bool InitializePoolIndex();
+bool DeinitializePoolIndex();
 
 } // end namespace glslang
 
--- a/dep/glslang/glslang/MachineIndependent/PoolAlloc.cpp
+++ b/dep/glslang/glslang/MachineIndependent/PoolAlloc.cpp
@@ -65,6 +65,15 @@ bool InitializePoolIndex()
     return true;
 }
 
+bool DeinitializePoolIndex()
+{
+    if (PoolIndex == OS_INVALID_TLS_INDEX)
+        return false;
+    OS_FreeTLSIndex(PoolIndex);
+    PoolIndex = OS_INVALID_TLS_INDEX;
+    return true;
+}
+
 //
 // Implement the functionality of the TPoolAllocator class, which
 // is documented in PoolAlloc.h.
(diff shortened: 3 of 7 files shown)


stenzek:
I was looking into this and was stumped why the TLS set would fail, that RA commit confirms my theory but doesn't really explain why.. Will have to remember to apply the patch when updating glslang.

My plans was to switch to shaderc or something else which wraps glslang eventually to reduce the binary size of the main executables, but that doesn't really help with the libretro core since it needs to be a single, self-contained binary.

The other option would be to ditch glslang completely and to ship SPIR-V. Unfortunately that relies on drivers behaving as expected with specialization constants, and I don't trust them to :)

jdgleaver:
> I was looking into this and was stumped why the TLS set would fail, that RA commit confirms my theory but doesn't really explain why.. Will have to remember to apply the patch when updating glslang.

Yes - Themaister (who authored that commit) is something of a legend in this field, and if he wasn't clear on the cause of the bug then I don't think *anyone* knows why it happens... :)

> My plans was to switch to shaderc or something else which wraps glslang eventually to reduce the binary size of the main executables, but that doesn't really help with the libretro core since it needs to be a single, self-contained binary.
>
> The other option would be to ditch glslang completely and to ship SPIR-V. Unfortunately that relies on drivers behaving as expected with specialization constants, and I don't trust them to :)

Ah, I'm just a humble RetroArch frontend developer, and most of these shader-related complications go over my head! But certainly, we can never rely on drivers to do anything correctly... (we've had *many* issues with GPU driver shenanigans, at least a couple of which remain completely unfixable on our end...)

Anyway, thanks for merging the PR! And thank you again for all your hard work. Duckstation offers an unparalleled combination of features and performance - the emulation community owes you a huge debt of gratitude!

PR #712, opened 2020-08-05
Only force 2 second pregap for track 1

### LINKED ISSUE #579: Rayman (NTSC-J) plays music out of order
It plays the wrong music for the menus and world map , when you finish the level and in some levels.
This is how the music should have sounded like in the first levels and on the menu : https://youtu.be/87GWSv3OFUM
Here is a picture of some debug info when the wrong music plays in the main menu:
image

RibShark:
Fixes #579

### DIFF
--- a/src/common/cd_image_cue.cpp
+++ b/src/common/cd_image_cue.cpp
@@ -132,7 +132,7 @@ bool CDImageCueSheet::OpenAndParse(const char* filename)
     // two seconds pregap for track 1 is assumed if not specified
     long pregap_frames = track_get_zero_pre(track);
     bool pregap_in_file = pregap_frames > 0 && track_start >= pregap_frames;
-    if (pregap_frames < 0)
+    if (track_num == 1 && pregap_frames < 0)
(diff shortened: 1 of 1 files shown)


stenzek:
I didn't notice that Rayman didn't have pregaps on tracks >2, thanks for catching.

Interestingly, the US and EU versions do have the pregap:
PAL: http://redump.org/disc/1055/
NTSC-U: http://redump.org/disc/8326/
NTSC-J: http://redump.org/disc/33719/

stenzek:
This will break bad dumps which don't explicitly include INDEX 0/the pregap, but in that case people should fix their dumps :)

RibShark:
> 
> 
> I didn't notice that Rayman didn't have pregaps on tracks >2, thanks for catching.
> 
> Interestingly, the US and EU versions do have the pregap:
> PAL: http://redump.org/disc/1055/
> NTSC-U: http://redump.org/disc/8326/
> NTSC-J: http://redump.org/disc/33719/

Yep, which is why those versions don't exhibit the same issue.

PR #741, opened 2020-08-11
Update gpusettingswidget.cpp

Abbanon:
Further line cleanup and word simplification in a few spots.


stenzek:
Thanks for the changes. I'll merge this when I get a chance to see if there's a way I can apply it to all languages rather than breaking all the translations.

PR #742, opened 2020-08-11
QtUtils: Improve QTableView column resizing

ggrtk:
This avoids use of hard-coded padding values and properly considers minimum and maximum column widths when calculating the remaining flexible width. Fixes unnecessary horizontal scroll bar showing up by default.

Before:
image

With patch:
image


stenzek:
Sounds good, thanks!

PR #749, opened 2020-08-12
Prefix libretro core options with core name.

dankcushions:
Until [only recently](https://github.com/libretro/RetroArch/commit/d9673b8c8fd86d8e2fdcfac503e3da2cd53af6e3) RetroArch's default behaviour was to save all core options for the installed cores in one global .cfg file. Now it defaults to splitting them out into individual files per-core, but the user may still configure it to use a global file via `global_core_options = "true"`. The latter is the behavior of my concern, the RetroPie project. 

To that end, it's suggested for libretro cores to prefix all core options with the name of the core, so they can be grouped and easily identified within a global .cfg. see https://github.com/libretro/RetroArch/blob/master/libretro-common/include/libretro.h#L1150-L1152

Eg:
```
beetle_psx_analog_calibration = "disabled"
beetle_psx_analog_toggle = "disabled"
beetle_psx_aspect_ratio = "corrected"
beetle_psx_cd_access_method = "async"
beetle_psx_cd_fastload = "2x(native)"
```
I have introduced the same behaviour here. Obviously any setting changes made before this change will be left as clutter in the file, and not retrieved. Could put some code in to pick up the previous settings (and subsequently save them under the new format) but you may consider that clutter in the code.

**NOTE:** I do see this in the logs:
```
[libretro WARN] [retro_set_environment] Failed to set core options, settings will not be changeable.
```
...however the new core options seem to save and load anyway :man_shrugging:. I might have missed something...

### DIFF
--- a/src/duckstation-libretro/libretro_host_interface.cpp
+++ b/src/duckstation-libretro/libretro_host_interface.cpp
@@ -150,7 +150,7 @@ std::string LibretroHostInterface::GetStringSettingValue(const char* section, co
                                                          const char* default_value /*= ""*/)
 {
   TinyString name;
-  name.Format("%s.%s", section, key);
+  name.Format("duckstation_%s.%s", section, key);
   retro_variable var{name, default_value};
   if (g_retro_environment_callback(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
     return var.value;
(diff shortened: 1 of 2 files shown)


stenzek:
Sounds reasonable, I'll trust your judgement here.

> NOTE: I do see this in the logs:
[libretro WARN] [retro_set_environment] Failed to set core options, settings will not be changeable.
...however the new core options seem to save and load anyway 🤷‍♂️. I might have missed something...

This is just me being pedantic and logging when the set options callback fails. RA calls set environment multiple times, and I have the options callback there. But options should only be set once, hence it fails?

PR #753, opened 2020-08-14
Hebrew Translation

MojoJojoDojo:


stenzek:
Thanks! Will hook it up the frontend.

PR #756, opened 2020-08-14
update zh translate

phoe-nix:


stenzek:
Thank you for your work! It's been added as part of #757.

PR #762, opened 2020-08-15
Update duckstation-qt_zh-cn.ts

phoe-nix:


stenzek:
Thanks for the update!

@ggrtk looks like there is some additional (new) strings translated in here. Do you think it's worth merging those with the current file in master?

PR #821, opened 2020-09-04
Fix D3D11 resource leaks

CookiePLMonster:
Fixes strange leaks in `AutoStagingTexture::EnsureSize` and `D3D11HostDisplay::DownloadTexture`. I say "strange" because in other places managed COM pointers were used, but those two were erroneously defined as naked pointers, as if the functions getting them did not increase their refcount.

`GetDevice` and `GetResource` both increase the refcount of the object they return, as described here:
https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicechild-getdevice
https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11view-getresource

### DIFF
--- a/src/common/d3d11/staging_texture.cpp
+++ b/src/common/d3d11/staging_texture.cpp
@@ -84,8 +84,8 @@ bool AutoStagingTexture::EnsureSize(ID3D11DeviceContext* context, u32 width, u32
   if (m_texture && m_width >= width && m_height >= height && m_format == format)
     return true;
 
-  ID3D11Device* device;
-  context->GetDevice(&device);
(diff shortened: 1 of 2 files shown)


stenzek:
I probably didn't realize that those functions increment the reference count.. that's what I get for being lazy and not checking the documentation :(

CookiePLMonster:
AFAIK with COM pointers every time you receive a new pointer from a function, it's expected to have its own reference - if it doesn't, then the interface implementation is broken*!

* Or D3D10, they thankfully rectified this idiocy in D3D11.

PR #826, opened 2020-09-05
Ass a 256x256 app icon

CookiePLMonster:
Windows can also use a 256x256 application icon for Large and Extra Large icons. Unlike smaller icons, this one is stored as a PNG so it doesn't explode the binary size.

Comparison between a new icon and an old icon, displayed at max icon size:
image

I have sourced the icon from a `APP_ICON_DATA` variable in code and cut it down from 260x260 to 256x256.


j8r:
Where is the source (SVG) of this icon? It is needed for #1987

CookiePLMonster:
> Where is the source (SVG) of this icon? It is needed for #1987

> I have sourced the icon from a APP_ICON_DATA variable in code and cut it down from 260x260 to 256x256.

It's embedded in the code.

stenzek:
SVG is not available. New icons for the emulator are welcome :)

j8r:
Thanks, I've read the code and found this URL: https://icons8.com/icon/74847/duck. Not sure about the license, even more because a signup is needed.

PR #838, opened 2020-09-10
Oh Update, here we go again! ;)

andercard0:
Translation Notes

Not translateable Texts list:

- Compiling Shaders;
- Resolution scale se to 16x;
- Keyboard/keypad+plus minus;
- OSD: Preloading Image to Ram;
- Restore Defaults / Close buttons;
- Options Load Game State 1 and so on;
- PGXP Enable/Disable Recompilling all Blocks;
- Tabs: General - Graphics - Save States and Audio;

Note 2: Wrong Description

- [ ] Automatically Load Cheats, when mouse over it; is wrongly showing General Settings instead of explain what the option is for.

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_pt-br.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_pt-br.ts
@@ -1016,6 +1016,359 @@
     </message>
 </context>
 <context>
+    <name>DisplaySettingsWidget</name>
+    <message>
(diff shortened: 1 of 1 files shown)


stenzek:
> Automatically Load Cheats, when mouse over it; is wrongly showing General Settings instead of explain what the option is for.

Probably because I forgot to add a description.

PR #844, opened 2020-09-12
Simplify quicksave/quickload logic

CookiePLMonster:
Pressing F1/F2 to save/load state for the first time would do nothing because of the list refreshing code. An obvious fix would be not to return after calling `RefreshList();` but upon closer examination, turns out the list is not required for quick saves/loads at all. This PR addresses the issue so the save state list is not populated upon save/load, and therefore fixes an aforementioned bug with the first save/load hotkey press.


stenzek:
Sounds good to me, thanks!

PR #848, opened 2020-09-13
More path fixes

CookiePLMonster:
XML loading and screenshot/texture saving **still** couldn't handle UTF-8 on Windows.

~~**NOTE:** For screenshot saving, I had to define `STBI_WINDOWS_UTF8` in multiple places (incl. stb project), is there a better way to do that? I probably also need to add the define to CMake somewhere?~~

### DIFF
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -1,4 +1,5 @@
 #include "image.h"
+#include "file_system.h"
 #include "log.h"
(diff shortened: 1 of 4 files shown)


CookiePLMonster:
Will be editing it later today.

CookiePLMonster:
Replaced the STBI commit with a better approach, as discussed on Discord.

PR #849, opened 2020-09-13
[libretro] add input descriptors

hunterk:
These are used by RetroArch to do input remapping. I used the same macro from pcsx_rearmed to avoid 7x duplicate sections (1 for each possible pad).


stenzek:
Thanks! Was meaning to do this.

PR #873, opened 2020-09-21
Russian translation

r57zone:
Hi, another translation for this convenient emulator.


stenzek:
Thanks, much appreciated! I'll hook this up in the frontend.

stenzek:
I've updated the OSD font renderer to use the Cyrillic glyph range when the Russian language is selected, so you should be able to translate the OSD messages now as well.

PR #880, opened 2020-09-23
update zh translation

phoe-nix:
"Auto-Detect" can not translate in "BIOSSettingsWidget".


stenzek:
Auto-detect should be translatable with https://github.com/stenzek/duckstation/commit/fb420bad82f39d6666b31f8494b48c55870f8d65

PR #903, opened 2020-09-26
CDImage fixes

### LINKED ISSUE #723: Preloading Image to RAM not compatible with LibCrypt
"Preloading Image to RAM" option is incompatible with PAL games with libCrypt copy protection.

For example, I encountered problems with Crash Team Racing PAL due to libCrypt (#646), which I solved by getting a .sbi file with the required sub channel data. I tried the option "Preload Image to RAM" and now I'm facing the same problem. Disableing it, solves the problem.

I am using a CHD file.

ggrtk:
Should resolve #723.

### DIFF
--- a/src/common/cd_image_bin.cpp
+++ b/src/common/cd_image_bin.cpp
@@ -25,17 +25,6 @@ private:
   CDSubChannelReplacement m_sbi;
 };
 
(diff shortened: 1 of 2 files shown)


stenzek:
Thanks!

Issue #861, opened 2020-09-17
Unable to run on Raspberry Pi OS (Debian "stable" Buster AARCH64)


vanfanel:
Hi @stenzek !

As you may remember, I was trying to cross-compile DuckStation as a LibRetro core, and I finally succeeded. The problem is that it can't be run:

```
i@raspberrypi:~/psx $ retroarch -v -L duckstation_libretro.so isos/Tekken\ 3/Tekken\ 3\ \(USA\).cue 
[INFO] [Overrides]: Redirecting save file to "/home/pi/.config/retroarch/saves/Tekken 3 (USA).srm".
[INFO] [Overrides]: Redirecting save state to "/home/pi/.config/retroarch/states/Tekken 3 (USA).state".
[INFO] === Build =======================================
[INFO] Capabilities:  ASIMD
[INFO] Built: Aug 24 2020
[INFO] Version: 1.9.0
[INFO] =================================================
[INFO] [CORE]: Loading dynamic libretro core from: "duckstation_libretro.so"
[ERROR] Failed to open libretro core: "/home/pi/psx/duckstation_libretro.so"
Error(s): /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/pi/psx/duckstation_libretro.so)
[INFO] Content ran for a total of: 00 hours, 00 minutes, 00 seconds.
[INFO] [CORE]: Unloading core symbols..
[INFO] [Video]: Does not have enough samples for monitor refresh rate estimation. Requires to run for at least 4096 frames.

```

As you can see, it seems to need LIBC 2.29. Currently, highest version in Debian Stable seems to be 2.28:

```
i@raspberrypi:~/psx $ ldd --version
ldd (Debian GLIBC 2.28-10) 2.28
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.

```

So, is that it? Is DuckStation incompatible with current Debian Stable (Debian Buster)?

Thanks!

dankcushions:
hmm, duckstation works fine for me on raspberry pi OS 64-bit (which is Debian buster), with the same version of LIBC as you:
```
pi@raspberrypi:~ $ ldd --version
ldd (Debian GLIBC 2.28-10) 2.28
```
verbose log:
```
Parameters: 
[...]

PR #908, opened 2020-09-29
gl/glcore context switching

barbudreadmon:
Preparatory patch for future merge of https://github.com/libretro/RetroArch/pull/11368
It might be a bit ugly so feel free to rewrite it to your own standard :).

### DIFF
--- a/src/duckstation-libretro/libretro_opengl_host_display.cpp
+++ b/src/duckstation-libretro/libretro_opengl_host_display.cpp
@@ -28,20 +28,31 @@ static bool TryDesktopVersions(retro_hw_render_callback* cb)
   static constexpr std::array<std::tuple<u32, u32>, 11> desktop_versions_to_try = {
     {/*{4, 6}, {4, 5}, {4, 4}, {4, 3}, {4, 2}, {4, 1}, {4, 0}, */ {3, 3}, {3, 2}, {3, 1}, {3, 0}}};
 
-  for (const auto& [major, minor] : desktop_versions_to_try)
+  retro_hw_context_type preferred_renderer;
+  if (!g_retro_environment_callback(RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, &preferred_renderer))
+    preferred_renderer = RETRO_HW_CONTEXT_DUMMY;
+
+  if (preferred_renderer == RETRO_HW_CONTEXT_DUMMY || preferred_renderer == RETRO_HW_CONTEXT_OPENGL_CORE)
   {
-    if (major > 3 || (major == 3 && minor >= 2))
+    for (const auto& [major, minor] : desktop_versions_to_try)
     {
-      cb->context_type = RETRO_HW_CONTEXT_OPENGL_CORE;
-      cb->version_major = major;
-      cb->version_minor = minor;
+      if (major > 3 || (major == 3 && minor >= 2))
+      {
+        cb->context_type = RETRO_HW_CONTEXT_OPENGL_CORE;
+        cb->version_major = major;
+        cb->version_minor = minor;
+
+        if (g_retro_environment_callback(RETRO_ENVIRONMENT_SET_HW_RENDER, cb))
+          return true;
+      }
     }
-    else
-    {
-      cb->context_type = RETRO_HW_CONTEXT_OPENGL;
-      cb->version_major = 0;
-      cb->version_minor = 0;
-    }
+  }
+
+  if (preferred_renderer == RETRO_HW_CONTEXT_DUMMY || preferred_renderer == RETRO_HW_CONTEXT_OPENGL)
+  {
+    cb->context_type = RETRO_HW_CONTEXT_OPENGL;
+    cb->version_major = 0;
+    cb->version_minor = 0;
 
     if (g_retro_environment_callback(RETRO_ENVIRONMENT_SET_HW_RENDER, cb))
       return true;


stenzek:
I'm... not sure why this is the core's responsibility. If core isn't supported, why is RetroArch returning success for the SET_HW_RENDER callback? Also it tries the versions in reverse order, so a core context will always be requested first.

GLES has the same problem, instead of returning false for an unsupported version, it quite happily accepts it and then crashes later on.

barbudreadmon:
Read https://github.com/libretro/RetroArch/pull/11368#issuecomment-700725090 to know what will happen with your core after the RA PR is merged if you don't merge (or rewrite) this.

And yes, context request versioning is unreliable, that's nothing new, lots of false positive and that's not always RA's fault, you should probably avoid asking for specific versions and use something like this later to check what version you actually received : 
```
int maj, min;
glGetIntegerv(GL_MAJOR_VERSION, &maj);
glGetIntegerv(GL_MINOR_VERSION, &min);
```

stenzek:
There is no way for the core to query the maximum supported version, so the core just had to try one until it works. This is also the standard way WGL, GLX, etc all work.

Retroarch crashing because the core is requesting an unsupported version is very much a retroarch fault.

Frankly I'd rather just *only* support glcore and gles. The OpenGL renderer is semi-deprecated in duckstation anyway, and a few things aren't emulated accurately unless you have texture buffers/SSBOs/dual source blend anyway, so there isn't much point in supporting <GL3.3.

stenzek:
To be clear, yes, I am very much aware you don't always get the version you ask for (I wrote the context wrappers in DS and Dolphin). But with GLX/WGL/EGL/etc, context creation will either succeed and give you a different version, or return an error.

Retroarch always succeeds, then crashes later on if the version isn't supported (e.g. rpi4 only supports gles3.1). We shouldn't be expected to target ancient hardware just because RA is broken.

barbudreadmon:
> Frankly I'd rather just only support glcore and gles. The OpenGL renderer is semi-deprecated in duckstation anyway, and a few things aren't emulated accurately unless you have texture buffers/SSBOs/dual source blend, so there isn't much point in supporting <GL3.3.

"gl" video driver also provide 3.3+ contexts. You can run several cores requiring modern gl with the "gl" video driver, but some won't or require additional settings (shared hardware context), that's what my RA PR is trying to solve by making a clear differentiation between "gl" and "glcore".

stenzek:
> "gl" video driver also provide 3.3+ contexts. You can run several cores requiring modern gl with the "gl" video driver, but some won't or require additional settings (shared hardware context), that's what my RA PR is trying to solve by making a clear differentiation between "gl" and "glcore".

Nothing should be relying on a non-core context providing anything past GL3.0 (or 3.1, I forget). I think the blob drivers on Windows are the only ones who even allow this, Mesa doesn't and IIRC Mac doesn't either.

I already have a check for the preferred renderer here https://github.com/stenzek/duckstation/blob/master/src/duckstation-libretro/libretro_host_interface.cpp#L963, which frankly is still too much - the core should just be able to query which APIs are supported and request one accordingly.

My argument is that if you're going to have a breaking change, you may as well fix the crashing at the same time, because that's a real bug people are experiencing (see https://github.com/stenzek/duckstation/issues/861) and the fact that you have to hack a fix in/set a macro and use a separate binary is ridiculous.

barbudreadmon:
> My argument is that if you're going to have a breaking change, you may as well fix the crashing at the same time

Well, retroarch will need a graphics engineer to fix those issues, i'm not. I'm just trying to fix an issue preventing several cores from working with default retroarch settings. Also, you mention an issue with raspberry pi os, well, knowing that pi4 gpu drivers are known for having several bugs requiring workarounds in various emulators, i wouldn't put the blame on retroarch without being sure.
Also, nvidia driver allows this on linux too : 
```
OpenGL core profile version string: 4.6.0 NVIDIA 450.66
OpenGL core profile shading language version string: 4.60 NVIDIA
OpenGL version string: 4.6.0 NVIDIA 450.66
OpenGL shading language version string: 4.60 NVIDIA
```

stenzek:
Yes, on the blob. Mesa only supports GL3 with a compat context.

> Also, you mention an issue with raspberry pi os, well, knowing that pi4 gpu drivers are known for having several bugs requiring workarounds in various emulators, i wouldn't put the blame on retroarch without being sure.

It absolutely is a retroarch issue. If the core requests a GLES3.2 context (like DuckStation), it can't create it and crashes, as you can see in the thread I linked. What it should do is try a GLES3.1 context, and so on, until it's successful, or track the maximum version in the first place and return an error when the core requests it, so it can explicitly request a lower version.

barbudreadmon:
That's a separate issue, and not something i can fix.

barbudreadmon:
also, don't misunderstand, i totally agree with you that "gl" video driver shouldn't be allowed with cores requiring modern gl, that was my first intent with those changes a few months ago, but i'm not the one who decide that (hence why my first attempt was reverted), that's why this time i've to deal with all those PRs to dozens of cores so that people can keep using "gl" video driver with core that requires opengl 3.3+.

barbudreadmon:
i just discussed this with @twinaphex, it's fine if you want to leave things as-is, just keep in mind that the behavior explained at https://github.com/libretro/RetroArch/pull/11368#issuecomment-700725090 will be expected, which is totally fine if you want duckstation to force the "glcore" video driver in the first place.

stenzek:
Have you actually tested DuckStation with the gl driver forced? If it can't get a hardware context, it shouldn't crash or refuse to boot, it'll just fall back to the software renderer.

But yeah, using glcore only would probably be the way to go. Less code.

barbudreadmon:
Using current duckstation with my RA PR :
- with "allow core to switch video driver" on, gl will always be replaced by glcore
- with "allow core to switch video driver" off, it fallbacks to gl after trying gcore 3.3 and 3.2

So yeah, no boot failure, at worse i guess if you leave things like this you'll only end up with some people asking how to force "gl" video driver

stenzek:
This seems to be working fine on master. As much I'd like to only use glcore, that cuts out people with ancient GPUs which only support GL3.0, and standalone works fine with it.

If you're forcing gl for some reason.. well, it'll still "work" I guess, depends on the driver as to whether you actually get a 4.x+ compatibility context.

PR #918, opened 2020-10-02
Fix stat64 failure on Apple Silicon

MaddTheSane:
The `stat64` struct seems to be designed to be a forward declaration under Apple Silicon according to the SDK in Xcode 12.2. A valid work-around is to just use the (not deprecated on OS X) `stat` (which is 64-bit since 10.6 via linking trickery, and always 64-bit on 64-bit OSes).


stenzek:
Neat! I'm a little surprised it compiled without any other issues. I'll be even more surprised if the JIT works... :)

MaddTheSane:
FYI: I’m trying to create an OpenEmu plug-in using the emulator as a base. I also don’t have a Developer Transition Kit, so I can’t test the JIT on Mac ARM.

PR #925, opened 2020-10-04
Qt: Add Make Portable action to Tools menu

ggrtk:


stenzek:
We'd probably want to move all the data files over too, otherwise people are going to complain "turning portable deleted my saves".

ggrtk:
Going to close this since it's outdated. Perhaps to avoid dealing with moving/copying user files and potential annoyances such as running out of disk space, we could query the user for selecting portable mode the first time the application is booted.

stenzek:
Hmm, it would have to be before anything is initialized then. Because the user directory setup is basically the first thing it does (and annoyingly, happens off the UI thread).

PR #928, opened 2020-10-04
Create update-and-edit-language_dragndrop.bat

RaydenX93:
Hopefully I'm making the PR correctly, I'm not a Git expert. Bear with me.

Anyhow, I created an helper .bat file for translators because I don't like setting an environment variable for the language I want to update.

You can drag and drop the .ts file on the .bat and it will automatically update it and then launch linguist.

References are included in the .bat file.


stenzek:
You need to add a bunch of other parameters to lupdate so it picks up the core strings. See here https://github.com/stenzek/duckstation/blob/b2d8af8463c583f94cc6b8514266f601869d5524/src/duckstation-qt/update_translations.bat#L1

RaydenX93:
Sure, I can do that.

However, do note that I just modified the old "update-and-edit-language.bat" in "src\duckstation-qt\translations" which does not appear to have that the other parameters.

Weird...

stenzek:
Sounds like I forgot to update it when I added the core strings :)

RaydenX93:
Let me know if it's good now.

Issue #362, opened 2020-04-28
UX/UI Improvements


stenzek:
Please feel free to add to this thread if you have any ideas which you think would make the UI better for you, as a user.

- [ ] Storing last played information with games so you can sort by most recent
- [x] Built-in memory card manager
- [x] Hide system console (#74)
- [ ] Grouping games into categories
- [ ] Automatic download of box art in game list
- [ ] Search feature for game list
- [ ] Displaying compressed size as well as uncompressed size in game list
- [x] Star rating for game compatibility (would need a web service/backend as well)
- [x] Discord API integration - displaying game name
- [x] Auto updater (#60)
- [x] Multiple mappings for buttons
- [ ] Launch random game from list
- [x] Integer scaling of display
- [ ] Disc image verification
- [x] Automatic analog mode enable for controllers
- [x] Saving of window size/position, column size/order
- [ ] Automatic renaming from game title<->game code memory cards
- [ ] Hide games from list
- [ ] Indicator when overrides are present
- [ ] Custom sort order/sort tags

Basmurd:
New feture, separate memcard per game, is awesome, but not intuitively for management. How about card name, have the name from tag cue, bin or iso file? Qt version not see iso files at all O_O

stenzek:
@Basmurd Could you clarify what you mean there? You can choose to have the memory cards named after the game code or the title of the game (which is the filename if you don't use the redump database). Or are you meaning pull some metadata from the cuesheet and use that?

Are there any PlayStation disc images in ISO format? Most games take advantage of CD/XA to some extent, so they need the raw 2352 byte sector dumps (bin format). That said, I could add a ISO image loader which dynamically generates the extra content if the use cases are there. Maybe some homebrew would use it?

Basmurd:
I watched the settings inattentively, such an option already exists. Sorry :) My other request is still valid. QT version does not see ISO files.

tony971:
* #75 can be merged into this
* support .m3u files for multi-disc games (and possibly replace the individual discs with them in the games list)
* use keyboard bindings if mapped controller isn't currently plugged

kurayami6:
The ability to mute the emulator would be nice.

ignacionelson:
@kurayami6 volume controls (up, down, mute) would be awesome

kurayami6:
[...]

PR #931, opened 2020-10-05
Add 16:10 Aspect Ratio Option

mirrornoir:
This time I didn't forget to declare it in `types.h` heh. This change was mostly so I could learn how to make a PR that changes two files, if this PR is unnecessary to due to proposed custom aspect ratio feature as mentioned here https://github.com/stenzek/duckstation/issues/362#issuecomment-686953114, then feel free to drop it.


Richard-L:
Are there games that use this, or will it just stretch the image to match 16:10?

mirrornoir:
This will just stretch the image like the rest of the Aspect Ratio options Richard-L, to fully benefit from an aspect ratio correct view there will need to be 16:10 specific widescreen cheats for use in combination with this setting. But we discussed this a bit in Discord.

PR #947, opened 2020-10-08
Zkdpower update chinese language ts.file 2020.10.9

zkdpower:
fix and improve the language duckstation-qt_zh-cn.ts file. unify every symbol and punctuation, idiomatically local translation


zkdpower:
only the final one is ok, previous version is useless.

stenzek:
I'm not sure what is going on with the PR here (says unknown repository) and the file was moved out of the translations folder.

So I just applied the patch manually, it is committed in https://github.com/stenzek/duckstation/commit/74880ac0474f7eb58a1647f78aa560a1060c7e6f

PR #958, opened 2020-10-11
Update for translation bat file

RaydenX93:
The script now allows for creation of a new translation.
Everything can be managed through my script, both creation and update.

@stenzek I believe that all the other bat files have been now superseded.
Maybe we should delete them? It's your call.

If so, I'll take care of updating the wiki as well.


stenzek:
Probably best to keep the others around in case anyone is using them. But feel free to update the wiki.

RaydenX93:
Got it!

TzakShrike:
This shouldn't be merged in because what I wrote already does this.

Issue #923, opened 2020-10-03
[Bug] Minor UI issues


hhromic:
I noticed two minor UI issues in Duckstation.

(1) At present, using the <kbd>Escape</kbd> key in the Settings and Memory Card Editor dialogs for quickly closing them seems to propagate the key back to the main window. This results in Duckstation trying to stop a currently running game after closing any of these dialogs (by default, <kbd>Escape</kbd> is the hotkey for "Power Off System").

(2) The Imgui debug dialogs have an "X" for closing them, however when using this, the corresponding toggle checkmark in the Debug menu is not updated. This results in Duckstation re-opening all previously closed debug dialogs that still have the checkmark when anything else is toggled in the Debug menu.

I know that the debug menu is not intended for regular users but wanted to make you aware of that minor glitch as well.

EDIT: Just noticed now that (1) is also true for the About and Updater dialogs. Probably all dialogs in the application are affected.

stenzek:
Second part is "fixed" by removing the close buttons from the windows in https://github.com/stenzek/duckstation/commit/c6c67b7b27ee752b159477cb81810aebc7e08880.

They were only there in the first place because the old SDL frontend would save the booleans directly to the config file, whereas Qt has always worked differently.

PR #989, opened 2020-10-20
CommonHostInterface: Activate hotkeys on press instead of release

ggrtk:
The volume up/down hotkeys were activating on press while the rest of the hotkeys were activating on release. This standardizes hotkeys to activate on press rather than release. Should also resolve the first part of #923 since closing dialogs with Esc triggers on a key down, but the hotkey bound to Esc would then be activated on the key up.


hhromic:
Can confirm that this PR solved for me the first part of the issue I reported 👍   Thanks!

Issue #369, opened 2020-04-28
BIOS screen doesn't fade out properly in King's Field (USA)


ghost:
Instead the text spazzes out and jumps around the screen. Happens with all renderers. 

![Capture](https://user-images.githubusercontent.com/31968468/80543504-563ab380-89f2-11ea-9312-e9bb5c7891ed.PNG)

stenzek:
This only happens for a couple of frames. Might be something to do with where the CRTC parameters are being latched.

ggrtk:
Fixed as of 6bc7cf2.

stenzek:
Fantastic!

PR #999, opened 2020-10-22
GPU: CRTC update

ggrtk:
Update to CRTC handling based on hardware tests.


ggrtk:
Latest commit should fix #369.

PR #1000, opened 2020-10-22
Core: Make some additional OSD messages translatable

ggrtk:


zkdpower:
some submenu under "debug" still shows in english, could you make it to be translatable?

stenzek:
Possibly. Depends how the strings are being constructed.

PR #1022, opened 2020-10-30
libretro: Fix duplicated core option keys

ggrtk:


stenzek:
Whoops, copy/paste fail, thanks for catching

PR #1039, opened 2020-11-03

zkdpower:


stenzek:
Added in https://github.com/stenzek/duckstation/commit/8e0ee7e9c5e061be734333932997144209579845

You need to hard reset your fork to the latest master to get rid of this extra file. If you're unsure of how to do that, deleting your fork and recreating it should also work.

PR #1045, opened 2020-11-05
AnalogController: Implement legacy rumble method

ggrtk:
Doesn't seem strictly necessary to save `m_legacy_rumble_unlocked` to state since it's only toggled during pad read sequences. I think the worst that could happen is missing or getting an extra frame of rumble on load state if the state was created very precisely in the middle of transferring the two motor bytes, so it's probably safe to wait for the next save state bump for saving the variable to state.


stenzek:
Thanks for this! Sorry for the delay in getting this merged, super busy at the moment. Will get to it as soon as possible.

RinMaru:
whats legacy rumble?

PR #1047, opened 2020-11-06
PS3's PS1 BIOS images

Blackbird88:
I just randomly tried if those work and they do! Not sure if useful to anyone so feel free to close this as further reference as I don't see any issues mentioning those 😀

### DIFF
--- a/src/core/bios.cpp
+++ b/src/core/bios.cpp
@@ -37,7 +37,7 @@ std::string Hash::ToString() const
   return str;
 }
 
-static constexpr std::array<ImageInfo, 25> s_image_infos = {{
+static constexpr std::array<ImageInfo, 27> s_image_infos = {{
   {"SCPH-1000, DTL-H1000 (v1.0)", ConsoleRegion::NTSC_J, MakeHashFromString("239665b1a3dade1b5a52c06338011044")},
   {"SCPH-1001, 5003, DTL-H1201, H3001 (v2.2 12-04-95 A)", ConsoleRegion::NTSC_U,
    MakeHashFromString("924e392ed05558ffdb115408c263dccf")},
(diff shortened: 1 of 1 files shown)


ggrtk:
I've been messing around with these a bit recently this week, but I haven't PRed support for them yet because 1. they cause a debug assertion failure in the debug builds due to unimplemented/oob EXP region writes (easy to squash, but it's better to at least properly stub the writes out to avoid assertion failure for this case) and 2. I think it might be best to confirm first what the EXP region writes are doing before officially recognizing these BIOSes. I know at the very least that the POST codes are being written to the first byte of EXP3 rather than the standard POST address at 0x1F802041.

I might be a bit overly cautious about this, but right now nothing's stopping users from manually setting these BIOSes and it's really up to stenzek if these should be formally recognized in DuckStation or not.

ggrtk (commented):
The PS3's PS1 BIOSes are not in SCPH-1000R, as far as I am aware.

Blackbird88:
I see. I thought they just got overlooked as PSP ones seem more attractive. My first thought was that more ways to get the BIOS legally is nice. Thanks for clarification!

ggrtk:
Working on some other stuff at the moment, but I'll take a further look into this over the weekend unless it's taken care of sooner.

ggrtk:
There's a hash in this PR that I chose not to add to the other one because it's for a ~4MB size PS2 BIOS found on earlier PS3 firmwares. DuckStation currently does not recognize non-512KB BIOSes, though it's simple enough to enable loading the PS1 BIOS out of the PS2 BIOS if there's enough user demand for that feature. There are an awful lot of PS2 BIOS revisions though, so I'm not sure if we really want to record the hashes for them anyway. Alternatively we could record hashes for the first 512KB of PS2 BIOSes if there are common PS1 BIOSes used across PS2s, but verifying that is not something I really care to spend time on.

Blackbird88:
Fair enough. I thought the other one was just earlier version.

PR #1050, opened 2020-11-08

zkdpower:


stenzek:
Can't see any differences between this and #1054 which is already merged. Please let me know if I have missed something.

PR #1072, opened 2020-11-14
Controller: Add Analog Joystick support

ggrtk:
Allows the use of analog controls for pre-Analog Controller titles such as Descent and MechWarrior 2.

### DIFF
--- a/src/core/analog_controller.cpp
+++ b/src/core/analog_controller.cpp
@@ -24,6 +24,7 @@ void AnalogController::Reset()
 {
   m_analog_mode = false;
   m_rumble_unlocked = false;
+  m_legacy_rumble_unlocked = false;
   m_configuration_mode = false;
   m_command_param = 0;
   m_motor_state.fill(0);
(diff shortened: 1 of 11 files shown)


src/core/analog_controller.cpp
```diff
@@ -41,8 +42,10 @@ bool AnalogController::DoState(StateWrapper& sw)
 
   sw.Do(&m_analog_mode);
   sw.Do(&m_rumble_unlocked);
+  sw.Do(&m_legacy_rumble_unlocked);
```
stenzek, line 45: This should be `sw.DoEx(&m_legacy_rumble_unlocked, 44, false);`

src/core/analog_controller.cpp
```diff
@@ -41,8 +42,10 @@ bool AnalogController::DoState(StateWrapper& sw)
...
   sw.Do(&m_configuration_mode);
   sw.Do(&m_command_param);
+  sw.Do(&m_button_state);
```
stenzek, line 48: Same here

stenzek (commented):
With the version check you don't need to put SAVEVERSION in the commit, since the old states should load fine.

ggrtk:
Didn't know about `DoEx`, thanks.

PR #1089, opened 2020-11-20

zkdpower:


zkdpower:
Current version has supported ISO image, so the next new version could add "*.iso" in "Start Disc" in frontend。

PR #1090, opened 2020-11-21
fix typo in italian translation

hugihadein:


RaydenX93:
Good catch!
Nice to see somebody actually using my translation! :D

PR #1108, opened 2020-11-26
AnalogController: Only clear motor state on rumble config if not used

ggrtk:
Works around low rumble in Crash Bandicoot Warped introduced in #1105.

When hit by an enemy while protected by Aku Aku, the game turns the small motor on, sends a rumble reconfiguration 2 frames later, and then turns the motor off another 2 frames after that. Not sure if a rumble reconfiguration forces motor state in any way, so this just restores the feel from before the previous PR.


stenzek:
Many thanks for chasing this down!

PR #1121, opened 2020-11-30
OES_geometry_shader and OES_copy_image support

dankcushions:
MESA is a little bit strange in that it incrementally exposes GL extensions before it reaches conformance for the particular GL versions. Eg, pi4 has several ES 3.2 extensions, but not all, so it's only ES 3.1 compliant. see:
[OES_copy_image](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/docs/features.txt#L280)
[OES_geometry_shader](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/docs/features.txt#L283)

This PR exposes the above two extensions, suppressing those warnings and (I hope?) permitting those code paths. However I'm not 100% sure what I've done is correct, and how to verify these extensions are working correctly. My test game (tekken 3) seems to run ok, and the warnings are no longer present.

I also think that something similar could be done for shader_cache support, maybe via the check here: https://github.com/stenzek/duckstation/blob/master/src/common/gl/shader_cache.cpp#L63 - `GL_NUM_PROGRAM_BINARY_FORMATS_OES` - but not sure about that one.

### DIFF
--- a/src/core/gpu_hw_opengl.cpp
+++ b/src/core/gpu_hw_opengl.cpp
@@ -208,8 +208,8 @@ void GPU_HW_OpenGL::SetCapabilities(HostDisplay* host_display)
   glGetIntegerv(GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT, reinterpret_cast<GLint*>(&m_uniform_buffer_alignment));
   Log_InfoPrintf("Uniform buffer offset alignment: %u", m_uniform_buffer_alignment);
 
-  if (!GLAD_GL_VERSION_4_3 && !GLAD_GL_EXT_copy_image)
-    Log_WarningPrintf("GL_EXT_copy_image missing, this may affect performance.");
+  if (!GLAD_GL_VERSION_4_3 && !GLAD_GL_EXT_copy_image && !GLAD_GL_ES_VERSION_3_2 && !GLAD_GL_OES_copy_image)
+    Log_WarningPrintf("GL_EXT/OES_copy_image missing, this may affect performance.");
 
 #ifdef __APPLE__
   // Partial texture buffer uploads appear to be broken in macOS's OpenGL driver.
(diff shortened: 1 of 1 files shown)


stenzek:
Sounds good, many thanks! I'll check out the program binary formats as that one's pretty critical - gets rid of the nasty compile time on startup. I'm assuming the Pi returns nonzero there?

For what it's worth, the geometry shader check isn't really necessary now, as I break lines into triangles rather than expanding them in the GS, but it's not hurting anything and another use could pop up in the future, so I kept it around.

dankcushions:
cheers!

> Sounds good, many thanks! I'll check out the program binary formats as that one's pretty critical - gets rid of the nasty compile time on startup. I'm assuming the Pi returns nonzero there?

hmm, I tried GL_NUM_PROGRAM_BINARY_FORMATS_OES: https://github.com/dankcushions/duckstation/commit/e9cdac5dd505a675ea7bdc198054d437d7a490ed
but it returns 0 still: 

```
[libretro INFO] [GL::ShaderCache] 0 program binary formats supported by driver
[libretro WARN] [Open] Your GL driver does not support program binaries. Hopefully it has a built-in cache$
```
however i thought it had shader binary storage implemented in other projects? eg https://github.com/gonetz/GLideN64/pull/1774
 
but maybe this is different from program binaries, which don't appear to be supported by v3d: https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/docs/features.txt#L146 - i'm not really sure what the deal is for GLES 3.x here.

stenzek:
`GL_NUM_PROGRAM_BINARY_FORMATS_OES` has the same value as `GL_NUM_PROGRAM_BINARY_FORMATS`, so that won't change anything.

As best I can tell, program binaries are not supported by v3d. For Mesa to return >0 formats, it needs `ctx->Const.NumProgramBinaryFormats` to be set to >0, which is done in gallium drivers in `mesa/state_tracker/st_extensions.c` on the condition of `get_disk_shader_cache` being not-null. Only the following drivers implement that callback:

```
$ grep -r "get_disk_shader_cache ="
gallium/drivers/iris/iris_screen.c:   pscreen->get_disk_shader_cache = iris_get_disk_shader_cache;
gallium/drivers/nouveau/nouveau_screen.c:   pscreen->get_disk_shader_cache = nouveau_screen_get_disk_shader_cache;
gallium/drivers/r600/r600_pipe_common.c:        rscreen->b.get_disk_shader_cache = r600_get_disk_shader_cache;
gallium/drivers/llvmpipe/lp_screen.c:   screen->base.get_disk_shader_cache = lp_get_disk_shader_cache;
gallium/drivers/etnaviv/etnaviv_screen.c:   pscreen->get_disk_shader_cache = etna_get_disk_shader_cache;
gallium/drivers/radeonsi/si_get.c:   sscreen->b.get_disk_shader_cache = si_get_disk_shader_cache;
gallium/drivers/tegra/tegra_screen.c:   screen->base.get_disk_shader_cache = tegra_screen_get_disk_shader_cache;
gallium/drivers/freedreno/freedreno_screen.c:   pscreen->get_disk_shader_cache = fd_get_disk_shader_cache;
gallium/drivers/r300/r300_screen.c:    r300screen->screen.get_disk_shader_cache = r300_get_disk_shader_cache;
gallium/auxiliary/driver_noop/noop_pipe.c:   screen->get_disk_shader_cache = noop_get_disk_shader_cache;
```

PR #1163, opened 2020-12-07
GPU: Fix GPUTexturePaletteReg y BitField width

### LINKED ISSUE #1162: [SLUS-01143] Ultimate Fighting Championship (USA)
**Game details**
[SLUS-01143] Ultimate Fighting Championship (USA)

**Description of the issue / bug**
Emulator crashes at second logo. Just happens with software renderer selected. With Hardware Renderer (Vulkan) the game runs normally.

**Steps to Reproduce**
Run the game with software renderer. The emulator crashes.

**Enhancements information**
Everything turned off.

**Hardware/software**
Windows 10 x64, Intel 4790, Geforce 1070

**Emulator version**
0.1-2475-g50273899

ggrtk:
Fixes #1162.

### DIFF
--- a/src/core/gpu_types.h
+++ b/src/core/gpu_types.h
@@ -195,7 +195,7 @@ union GPUTexturePaletteReg
   u16 bits;
 
   BitField<u16, u16, 0, 6> x;
-  BitField<u16, u16, 6, 10> y;
+  BitField<u16, u16, 6, 9> y;
 
   ALWAYS_INLINE u32 GetXBase() const { return static_cast<u32>(x) * 16u; }
   ALWAYS_INLINE u32 GetYBase() const { return static_cast<u32>(y); }


stenzek:
Strange, the bitfield class should do the masking/shifting. Will have to check why that's not happening.

src/core/gpu_types.h
```diff
@@ -198,7 +198,7 @@ union GPUTexturePaletteReg
   BitField<u16, u16, 6, 10> y;
```
stenzek, line 198: Does changing this from `BitField<u16, u16, 6, 10> y` to `BitField<u16, u16, 6, 9> y` also fix it? The code should be the same as the mask/shift.

ggrtk:
https://github.com/stenzek/duckstation/blob/50273899c66a998475fd2c7154349e203a57a9b9/src/core/gpu_sw.cpp#L663-L668

It's not getting masked out here when dispatching rectangle render commands for the software renderer. Setting the bit count for y to 9 works or we could apply the mask in the snippet above, whichever you prefer.

stenzek:
I'd prefer setting it to 9, since the fact that I have it listed as 10 bits there is clearly incorrect in the first place :)

PR #1170, opened 2020-12-08
Android - String Translation

andercard0:
Android. Strings translated for the latest Brazilian Portuguese


stenzek:
Added in https://github.com/stenzek/duckstation/commit/5c2efd880d0d40dd82aa86955a0dca5567d88b33

PR #1171, opened 2020-12-08
Android - Arrays Translation

andercard0:
Translated Arrays for Brazilian Portuguese


stenzek:
Added in Added in https://github.com/stenzek/duckstation/commit/5c2efd880d0d40dd82aa86955a0dca5567d88b33

PR #1173, opened 2020-12-08
First Android Translation

RaydenX93:
This is my first attempt at the Android version translations. Typo fixes, if any, will follow.

I'm don't think that my translation is supposed to overwrite the original files, so for now I created the "_it" versions.
If/when merged, feel free to put them where you see fit, @stenzek .


stenzek:
Added in https://github.com/stenzek/duckstation/commit/25fb3c01c339b4fcc85ccbefdfe84db8803518d7 - you can update the files in the `values-it` directory now. Sorry about the confusion with the strings, most weren't resources in the previous master version and couldn't be translated, they're all listed in `values/strings.xml` now.

PR #1174, opened 2020-12-08
AnalogController: Disallow analog-to-dpad in config mode

ggrtk:
Since pad reads in config mode get the full analog mode response regardless of actual digital/analog mode, this feature should be disabled in config mode. Don't remember at the moment which games do pad reads in config mode to test this, but this change should be safe.


stenzek:
Sounds reasonable, thanks!

PR #1192, opened 2020-12-10
GDB server stub for remote debugging

boricj:
I'm currently in the process of reverse-engineering a PS1 game and I need a good, modern PS1 emulator with a decent debugger. This is surprisingly hard to come by, so I've settled for a good, modern PS1 emulator and stick a GDB stub in it.

To use this, you need a GDB with MIPS support (I use gdb-multiarch with `set architecture mips:3000`). Enable the GDB server in the settings file, start the Qt frontend, boot something and connect to the server with `target remote :1234`.

Settings to enable the GDB server:
```
[Debug]
EnableGDBServer = true
GDBServerPort = 1234
```

This is extremely bare-bones at the moment, but it can peek/poke memory. My main concern so far is handling instruction-level breakpoints/watchpoints, since it appears the emulator can only be paused at frame intervals (which would be a couple million instructions too late).

TODO:
- [X] GDB server settings in configuration file
- [ ] GDB server settings in Qt UI
- [X] GDB server for Qt frontend
- [ ] GDB server for SDL frontend
- [X] Basic GDB server infrastructure
- [X] Read/write memory
- [X] Read/write general registers
- [X] Hackish frame-level interrupt/continue
- [x] Proper instruction-level interrupt/continue
- [ ] Instruction step
- [x] Breakpoints
- [ ] Watchpoints

### DIFF
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -163,6 +163,33 @@ std::size_t Strlcpy(char* dst, const std::string_view& src, std::size_t size)
   return len;
 }
 
+std::optional<std::vector<u8>> DecodeHex(const std::string_view& in)
+{
+  std::vector<u8> data;
+  data.reserve(in.size()/2);
+
+  for (int i = 0; i < in.size()-1; i+=2) {
+    auto byte = StringUtil::FromChars<u8>(in.substr(i*2, 2));
+    if (byte) {
+      data.push_back(*byte);
+    }
+    else {
+      return { };
+    }
+  }
+
+  return { data };
+}
+
+std::string EncodeHex(const u8* data, int length)
+{
+  std::stringstream ss;
+  for (int i = 0; i < length; i++) {
+    ss << std::hex << std::setfill('0') << std::setw(2) << static_cast<int>(data[i]);
+  }
+  return ss.str();
+}
+
 #ifdef WIN32
 
 std::wstring UTF8StringToWideString(const std::string_view& str)
--- a/src/common/string_util.h
+++ b/src/common/string_util.h
@@ -3,9 +3,11 @@
 #include <cstdarg>
 #include <cstddef>
 #include <cstring>
+#include <iomanip>
 #include <optional>
 #include <string>
 #include <string_view>
+#include <vector>
 
 #if defined(__has_include) && __has_include(<charconv>)
 #include <charconv>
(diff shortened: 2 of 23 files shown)


ggrtk:
Not sure if you've seen it, but there's a WIP debugger on the `debugger` branch of the repo.

boricj:
@ggrtk I haven't. But it seems incomplete and it does not compile...

The two versions don't have much in common : one is an integrated GUI debugger written in Qt, the other is a headless TCP server implementing the GDB remote protocol specification. A GDB server would enable advanced features like scripting, expression evaluation including function calls and debugging symbols out of the box since all of that is implemented inside the client, whereas I'd expect an integrated debugger to be much more limited because it can't rely on GDB as a debugging engine and would need to implement each feature by itself. An integrated debugger would be self-contained and more user-friendly.

That being said, they could share debugger hooks and they should be able to co-exist, so there's no harm in having both.

Comparing the debugger hooks, my peek/poke approach through the bus should work for every case (RAM, ROM, hardware device, unmapped memory, cache invalidation...) whereas exposing memory regions will only work with ROM and RAM, but would be much faster to scan/search through.

stenzek:
Hi there! Thanks for taking the time to put this together.

A few notes:
 - Do you want to be able to peek/poke MMIO addresses? Some of these will call out to C++ handlers (and potentially the graphics driver), which will race or crash on a different thread.
 - Therefore, it may be better to poll the gdb connection on the main thread.
 - Using BSD sockets instead of Qt would allow the debugger to sit in the core instead of the frontend, and reduce code duplication. I can potentially switch this over later if needed.

For breakpoints, my plan was to just force the interpreter when using the debugger. It's still fast enough to use, and adding breakpoints in the recompiler complicates things a nontrivial amount. Of course it's possible to do, but I'm just not sure if it's worthwhile :)

boricj:
I don't intend to peek/poke MMIO addresses since I want to reverse-engineer a PS1 game and not Psy-Q, but someone else might want to. I'm not familiar with duckstation's internals nor Qt so I can't really comment on the rest for now. I'll study the emulator a bit more to better understand how everything could fit together.

stenzek:
Sure :) Feel free to hit me with questions if needed here or on discord.

boricj:
I've rebased my GDB server on top of the `debugger` branch and it seems to work just fine (https://github.com/boricj/duckstation/tree/gdbserver-with-debugger).

@stenzek Any ETA on a stable-ish base commit I could rebase this on for this pull request? It only need the core debugging infrastructure and not the integrated debugger UI, so splitting out the core part of the `debugger` branch into a separate base commit should do the trick.

stenzek:
Will try to get the debugger merged into master in the next couple of days, it's mostly finished, just have to do some regression testing to make sure I haven't broken anything.

Issue #706, opened 2020-08-04
[Libretro] Auto analog mode not working


ds22x:
Don't know if the issue is already known, but the auto analog mode option currently doesn't do anything for the Libretro core.

stenzek:
You have the controller type set to analog, right?

ds22x:
Yes I have, but it doesn't seem to automatically switch.
It's especially noticable with games that refuse to work with the DualShock.

stenzek:
Pretty sure I did check it a while back.

What do you mean by "games that refuse to work with dualshock"? If you have the auto enable option enabled, those games shouldn't have any controller input at all (it won't be recognized).

ds22x:
That's right, when auto is enabled and the controller type is set to analog, those games don't have any inputs at all.
But yeah, the auto option doesn't seem to work, neither when the control type is set to analog nor digital.

stenzek:
That's... working as intended then?

It'll have no effect with the type set to digital controller - `Automatically enables analog mode in supported controllers at start/reset.`. The only supported controller for this is the analog controller.

If it wasn't auto-switching to analog mode, those games which aren't dualshock compatible would receive d-pad input as usual. This is the reason why I was very hesitant to add the option in the first place, but people kept asking for it.

ds22x:
I see.
I assumed it would work the same way Beetle PSX handles it, where it detects when a game supports the DualShock, it will switch the controls to analog, and keeps them digital when it doesn't support them after the BIOS.

ds22x:
E.G.:
https://youtu.be/HV-Y9tQqrm4
https://youtu.be/NpZot5Xbe9M

ggrtk:
> I see.
> I assumed it would work the same way Beetle PSX handles it, where it detects when a game supports the DualShock, it will switch the controls to analog, and keeps them digital when it doesn't support them after the BIOS (an example of this is with Rayman, where I have to specifically set it to digital with DuckStation for the controls to work, whereas with Beetle leaving the control type to DualShock and enabling the Analog toggle will still allow for functioning controls).

Beetle doesn't do any sort of automatic detection. Beetle's "Enable DualShock Analog Mode Toggle", like the description says, allows for the emulated DualShock controller to be toggled between digital mode and analog mode like original hardware. When this option is disabled, the DualShock controller is forced to analog mode and you can't toggle modes at all.

[...]

PR #1195, opened 2020-12-11
AnalogController: Clarify option for forcing analog mode

ggrtk:
The phrase "Auto Analog Mode" could imply to users that the option allows the game to automatically switch the controller to analog mode when supported (i.e. config command functionality), which is not what this option does. Changing the wording to "Force Analog Mode on Reset" should hopefully clear up the confusion in #706.


stenzek:
Android needs to be updated as well with the new variable name, but I'll change this.

PR #1224, opened 2020-12-14
Cccalibrator

PavelDurov1488:
Its not complete, i need to add dynamic noise and blur and make saturation tunable

### DIFF
--- a/data/shaders/Cccalibrator.glsl
+++ b/data/shaders/Cccalibrator.glsl
@@ -1,0 +1,114 @@
+/*==========================================*\
+/*=======Calibrator by PavelDurov1488=======*\
+/*==========It's not ready yet...===========*\
+/*===I need to add blur, dynamic noise...===*\
+[configuration]
+
+
+[OptionRangeFloat]
+GUIName = Luminance(Y)
+OptionName = LUMINANCE
+MinValue = 0.000
+MaxValue = 3.000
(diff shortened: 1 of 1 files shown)


data/shaders/Cccalibrator.glsl
```diff
@@ -0,0 +1,114 @@
...
+
+[/configuration]
+*/
+
+float pseudoNoise(vec2 co)
+{
+return fract(sin(dot(vec2(co.x+0.513,co.y+0.4124) ,vec2(12.9898,78.233))) * 43758.5453);// *fract(sin(dot(vec2(co.x+4.231,co.y+3.143) ,vec2(12.9898,78.233)*2.0)) * 43758.5453); //pseudo random number generator
+}
+
+const vec3 RGBtoY = vec3(0.299, 0.587, 0.114);
```
stenzek, line 72: You'll probably want to use `CONSTANT` here instead of `const`, otherwise it won't compile as HLSL.

For what it's worth, I'll be moving to using spirv-cross for compiling shaders in the updated system, so this won't be needed after that.

PavelDurov1488: In D3D11 it works, actually its not used...
I adapted simple-brightness shader to do RGB->YIQ->RGB conversation without this :D

Issue #1138, opened 2020-12-03
Crop window which preserves aspect ratio


Awakened0:
**Description of the issue / bug**
Changing the new "Display Line Offset" options stretches the aspect ratio instead of retaining it.

**Steps to Reproduce**
Set the Start offset to 12 and the End one to -12 in Crash 1. The image will be stetched vertically compared to 0 offsets:

![Crash Bandicoot (USA)-201203-172057](https://user-images.githubusercontent.com/4565863/101096549-b51a6980-358d-11eb-93e4-4fab1691bcae.png)

![Crash Bandicoot (USA)-201203-172300](https://user-images.githubusercontent.com/4565863/101096567-b8adf080-358d-11eb-85eb-054396c50b61.png)

**Hardware/software**
Tested the libretro core in both Software and Vulkan renderer modes.

**Emulator version**
0.1-2433-g9089c973

stenzek:
This is working as intended. What you're asking for is a crop offset which will preserve the aspect ratio.

ggrtk:
This is something I had in mind with an "auto" aspect ratio in #654.

Awakened0:
Both 1:1 PAR and the new Auto ratio now adjust properly with the offsets. Thanks!

Issue #654, opened 2020-07-23
Configurable overscan and auto aspect ratio


ggrtk:
We could consider adding this in functionality eventually (could be exposed to the user or not, either way) as it would be something we can gameini, as discussed in #596. Would also need to add an "auto" aspect ratio to compensate for some cases, since the aspect ratio is currently applied as fixed values from the host display side.

ggrtk:
Closing as #1227 has been merged.

PR #1227, opened 2020-12-15
GPU: Add auto aspect ratio and various fixes

### LINKED ISSUE #1139: 4:3 Aspect Ratio Identical to 1:1 PAR
**Description of the issue / bug**
The 4:3 aspect ratio option should be a little wider than 1:1 PAR, but currently they are the same (at least in the libretro core). I can achieve a proper 4:3 aspect by using the "All Borders" Crop Mode, which crops the vertical bars and stretches the image to 4:3. But ideally cropping options shouldn't change the aspect ratio.

**Steps to Reproduce**
Change the Aspect Ratio option to 4:3. It will have square pixels, just like the 1:1 PAR mode.

**Hardware/software**
Tested the libretro core in both Software and Vulkan renderer modes.

**Emulator version**
0.1-2433-g9089c973

ggrtk:
Auto mode width is roughly 2% slimmer than 4:3 mode due to using GPU ticks/lines for calculation. We could probably fudge the numbers a bit to make it closer to forcing 4:3, but analog specs aren't super precise anyways so I won't bother with it for the time being.

Should resolve #1139, #1138, #654.


stenzek:
Would "Auto (4:3)" be a better name? Auto might be confused for being based on the window size.

ggrtk:
How does "Auto (Native)" sound? Since it's not necessarily maintaining a 4:3 display aspect ratio depending on the user's crop settings.

stenzek:
Hm, "Native" still sounds like it could be confused for the host. How about "Auto (Game)"? (open to other suggestions of course)

DonelBueno:
Auto (Original)?

ggrtk:
Maybe "Auto (Game Native)" or "Auto (Native to Game)"? Gets us pretty specific about it without being too wordy, I think.

stenzek:
`Auto (Game Native)` works for me.

ggrtk:
Done -- not sure how to expose this setting for the Android build, so if that could be handled separately it would be greatly appreciated. Thanks for the feedback!

stenzek:
Great :) Will sort out Android and merge tomorrow with the other more significant changes.

Awakened0:
With this pull, will only the Auto 4:3 aspect properly adjust when cropping is applied? Or will the other aspect options, particularly 1:1 PAR, also adjust to your cropping settings?

ggrtk:
PAR 1:1 adjusts the aspect ratio for cropping adjustments because it's based on the visible resolution of the VRAM rectangle, which cropping affects. This is how it has been in standalone, and this behavior should be reflected in the libretro core as well.

PR #1244, opened 2020-12-17
GDB server stub for remote debugging (using upstream debugging interfaces)

boricj:
Just like #1192, but rebased on top of upstream's debugging interfaces and without watchpoints.

### DIFF
--- a/src/common/string_util.h
+++ b/src/common/string_util.h
@@ -3,9 +3,11 @@
 #include <cstdarg>
 #include <cstddef>
 #include <cstring>
+#include <iomanip>
 #include <optional>
 #include <string>
 #include <string_view>
+#include <vector>
 
 #if defined(__has_include) && __has_include(<charconv>)
 #include <charconv>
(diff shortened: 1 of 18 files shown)


src/duckstation-qt/mainwindow.cpp
```diff
@@ -814,6 +815,16 @@ void MainWindow::updateEmulationActions(bool starting, bool running)
     }
   }
 
+  if (g_settings.debugging.enable_gdb_server) {
```
stenzek, line 818: Probably not the best place for this, but we can move it later if we move the server to the core.

src/common/string_util.h
```diff
@@ -90,22 +113,26 @@ inline std::optional<bool> FromChars(const std::string_view& str)
 
   return std::nullopt;
 }
-
+#if 0
```
stenzek, line 116: If this is replaced by the above implementation, can just remove it rather than `#if 0`'ing it out.

src/duckstation-qt/gdbconnection.cpp
```diff
@@ -0,0 +1,82 @@
...
+  qint64 bytesRead;
+  char buffer[256];
+
+  while ((bytesRead = m_socket.read(buffer, sizeof(buffer))) > 0) {
+    for (char c : std::string_view(buffer, bytesRead)) {
+      m_readBuffer.push_back(c);
+
+      if (GDBProtocol::IsPacketInterrupt(m_readBuffer)) {
+        Log_DebugPrintf("(%u) > Interrupt request", m_descriptor);
+        g_host_interface->PauseSystem(true);
```
stenzek, line 40: Same here - this should use `pauseSystem`. Specifically `pauseSystem(true, true)`, that way it blocks the server thread until the emu thread actually pauses, and doesn't race when executing other commands such as adding breakpoints.

src/duckstation-qt/gdbconnection.cpp
```diff
@@ -0,0 +1,82 @@
...
+        m_readBuffer.erase();
+      }
+      else if (GDBProtocol::IsPacketContinue(m_readBuffer)) {
+        Log_DebugPrintf("(%u) > Continue request", m_descriptor);
+        g_host_interface->PauseSystem(false);
```
stenzek, line 45: `pauseSystem(false, false)` should be okay here I think, as the client won't send any other commands before another interrupt request?

stenzek (commented):
Thanks again for working through this.

I have some threading concerns, mainly with the commands - if they execute only after sending an interrupt first, replacing those `PauseSystem` calls will do the trick. But if breakpoints can be added during execution, you'll probably want to use `executeOnEmulationThread` if the system isn't paused.

I still think we're better off implementing this in the core rather than Qt, but that conversion can happen later, or I can do it myself :) But otherwise, looks good!

PR #1254, opened 2020-12-18
Update bug report issue template

ggrtk:


stenzek:
FWIW, https://github.com/stenzek/duckstation/commit/9e1987063d12995b86bd5f7de1f9fcc699e1e462 should hopefully solve these libcrypt issues being posted, but users can click through and ignore it.

PR #1256, opened 2020-12-18
 Added new cheat codes types A6, A7, F0, F1, F2 & F3

PugsyMAME:
Added some new unique cheat types, format information will be added to the chtdb.txt header on my next update.

Type A7 still needs fixing to get the Restore functionality to work   
    ExtConstantWriteIfMatch16 = 0xA6,
    ExtConstantWriteIfMatchWithRestore16 = 0xA7,
    ExtConstantForceRange8 = 0xF0,
    ExtConstantForceRangeLimits16 = 0xF1,
    ExtConstantForceRangeRollRound16 = 0xF2,
    ExtConstantForceRange16 = 0xF3.


stenzek:
I cherry-picked the cheat update commit, so that's in the latest build.

I was thinking instead of the global, we could have another function which iterates the instructions (`DisableCode` or something), which does the restoring. That way if you had a restore instruction behind a conditional, it'll still be restored even if the comparison fails.

If you think this sounds reasonable, I'm happy to make the changes, or you can - I don't mind either way :)

PugsyMAME:
Thanks.

I originally attempted a sort of DisableCode function which didn't work as I didn't understand the code, so I went for the hacky global solution to try and get an idea of a better solution but that also didn't work. I decided I best leave it to the expert - there would be too much of a danger that it would turn out to be a hackfest if I had carried on.... 

I'm hoping to look at your solution to learn from!

PugsyMAME:
Added some more cheat types for BIT cheats....that should do fo a while :)

https://github.com/PugsyMAME/duckstation/commit/47b69190d3fccd5078ba4b9b99dc206903622cd3

stenzek:
I've added this in https://github.com/stenzek/duckstation/commit/edbfc5487a04788ee31aecb0a5bad48b4afc101c, please let me know if I missed anything or if it does not work as intended.

PugsyMAME:
A7 code doesn't work as hoped, it enables fine but doesn't poke the value back on disable.
Example code for 40 Winks (USA) (1999) (GT Interactive Software) {SLUS-00874} <40winks>:
#Multi-Jump
A702B312 10401400

When the cheat enables it pokes 8002B312 with 1400 - which it does. As you will be able to see if you keep pressing fire on the first screen you will effectively float on the ceiling of the bedroom.
If you turn the cheat off it should check that 8002B312 contains 1400 and if it does (which it does) it will poke it with 1040 - which it doesn't do.

I've not looked at it yet, it may be something I missed on the original PR.

PugsyMAME:
All working now, whatever you did....or I didn't do. :)

PR #1260, opened 2020-12-19
Pending

VermillionFury007:
Make the emulator portable.
merge settings.ini folder(C:\Users\robot\Documents\DuckStation) with .exe folder


stenzek:
https://github.com/stenzek/duckstation#user-directories

> If you wish to use a "portable" build, where the user directory is the same as where the executable is located, create an empty file named portable.txt in the same directory as the DuckStation executable.

PR #1266, opened 2020-12-19
Use proper 21:9 aspect ratio, add 32:9

Margen67:
https://en.wikipedia.org/wiki/Ultrawide_formats#Ultrawide_Cinema


stenzek:
Thanks! The GTE should probably be updated to include these new ratios, so it works with the widescreen hack.

https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L620 and https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L715

I believe this should be `(x * 9) / 16` for 21:9 (64:27), and `(x * 3) / 8` for 32:9.

Margen67:
> 
> 
> Thanks! The GTE should probably be updated to include these new ratios, so it works with the widescreen hack.
> 
> https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L620 and https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L715
> 
> I believe this should be `(x * 9) / 16` for 21:9 (64:27), and `(x * 3) / 8` for 32:9.

@stenzek What should the Sx values be?

Margen67:
> 
> 
> Thanks! The GTE should probably be updated to include these new ratios, so it works with the widescreen hack.
> 
> https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L620 and https://github.com/stenzek/duckstation/blob/master/src/core/gte.cpp#L715
> 
> I believe this should be `(x * 9) / 16` for 21:9 (64:27), and `(x * 3) / 8` for 32:9.

Done.

stenzek:
Thanks! The game list cache version will need to be changed because the array indices for some of the ratios will be different now, but I'll take care of that :)

PR #1284, opened 2020-12-22

zkdpower:


zkdpower:
Cpu debugger interface should add  translation "register, valve" in the register column and "address ,valve" in the stack column.

PR #1295, opened 2020-12-23
Delete appveyor.yml

Margen67:


stenzek:
I keep this around in case people are still grabbing builds from AppVeyor, and it doesn't really hurt anything. (AppVeyor was used before I switched to GitHub Actions)

PR #1303, opened 2020-12-25
Fix a few path manipulation bugs

michaelforney:
I noticed these when trying to load a file from the current directory. When I tried `file.m3u`, I got a  `std::bad_alloc` exception, and when I tried `./file.m3u`, it tried to load items from `/`.


stenzek:
Nice catch, thanks! This code dates back many years, and obviously I wasn't considering this case :)

PR #1304, opened 2020-12-25
Fix build with musl libc

michaelforney:
Here are a couple of changes I needed to build on Linux with musl libc.


stenzek:
Great! :)

PR #1316, opened 2020-12-26
Changed debugger "Enter memory address" to accept hex only

PugsyMAME:
The "Enter memory address" prompt by default expects a decimal address unless it's preceded by 0x. Or it expects an number starting with 0 is an octal.
The disassembly address should be hexadecimal regardless as that is how it it displays the address.
Also changed it so that it changes any address entered to be divisible by 4 as there was an observed issue that would cause the disassembly addresses to get locked to a address that was not divisible by 4


PugsyMAME:
Replaced, updated

Issue #1026, opened 2020-11-01
OpenGL SSAA


KrossX:
Running on W7 + HD5770, so using deprecated hardware with deprecated renderer on deprecated OS (=P). That aside, there's no difference when I use SSAA over MSAA even though the log indicates it to be working. Adding the following code at the bottom of [GPU_HW_OpenGL::UpdateSettings()](https://github.com/stenzek/duckstation/blob/b5b1c725c1a561dc95e97284add4706884d76242/src/core/gpu_hw_opengl.cpp#L119) made it work as expected, like with D3D11.

```cpp
  if (m_multisamples)
    glEnable(GL_MULTISAMPLE);
  else
    glDisable(GL_MULTISAMPLE);

  if (m_per_sample_shading)
  {
      glEnable(GL_SAMPLE_SHADING);
      glMinSampleShading(1.0f);
  }
  else
    glDisable(GL_SAMPLE_SHADING);
```

stenzek:
Sounds like the driver's not following the spec then. GL_MULISAMPLE shouldn't be required, nor should GL_SAMPLE_SHADING (the inputs to the shader enable sample rate shading).

KrossX:
Driver bug, setting `m_use_glsl_interface_blocks` to false for the shader generator makes it work as well.

I couldn't find the default values for those enums so I checked with the reference book and indeed, GL_MULTISAMPLE is for the render to window side and defaults to enabled anway. GL_SAMPLE_SHADING sets `in` variables as `sample` by default, in case you cannot edit the shader.

PR #1328, opened 2020-12-27
Hack: Shader Interface Block INI setting

KrossX:
As found in #1026, old Radeon OGL driver has issues with the qualifier used for SSAA and Interface Blocks. This PR adds an INI setting that allows to disable shader Interface Blocks from being used and avoid the bug.

Currently, I've been just making custom builds with the interface blocks disabled so I decided to try and make a setting for it. It's the least invasive way I found, although it exposes global settings to the shader generator. No UI settings either, too hacky for "advanced" even.

### DIFF
--- a/src/core/host_interface.cpp
+++ b/src/core/host_interface.cpp
@@ -507,6 +507,7 @@ void HostInterface::SetDefaultSettings(SettingsInterface& si)
   si.SetIntValue("GPU", "ResolutionScale", 1);
   si.SetIntValue("GPU", "Multisamples", 1);
   si.SetBoolValue("GPU", "UseDebugDevice", false);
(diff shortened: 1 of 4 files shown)


stenzek:
Given this only affects old radeon drivers, wouldn't it make more sense to just test for that driver rather than exposing an option?

Can check GL_RENDERER.

KrossX:
Well, I know it happens with this one mix of old card/driver. I do not know if it doesn't happen with newer cards/drivers. Even if people were affected though, I doubt they would notice.

I have a test similar to the anisotropic test I made before to check for a bug on newer cards/drivers. It can already show the problem, but has no text at all yet. Maybe with that I'll know if it's been fixed.

PR #1344, opened 2020-12-29
Use gtk3 for nativefiledialog

guibou:
Rational: gtk2 is not maintained anymore and the change was simple.

- Update of nativefiledialog/CMakeLists.txt in order to detect and use
  gtk3
- Updated a few files referencing gtk2, especially:
  - github workflow: I tested that the apt-get update works on ubuntu
    18.03 in a docker container
  - Readme.

This was tested in the following nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/107878


stenzek:
Sounds reasonable, many thanks!

stenzek:
Btw, if you're packaging, you may want to track the `dev` branch rather than `master` - I have two releases, `master` is bleeding edge and may have regressions, which I try to fix before pushing to `dev`. Usually they're only a few days apart :)

guibou:
@stenzek Good to know. Thank you.

Sunderland93:
Why not using Qt?

stenzek:
Because Qt is a huge dependency to pull in for just a file selector.

guibou:
@Sunderland93 I had the same interrogation and I realized that nativefiledialog is only used with the sdl UI. So if you only install that one, you pull a small dependency (`gtk`).

On the other hand, if you package both UI at the same time (that's what I've done for nixpkgs), then you pull both UI libraries ;)

stenzek:
My suggestion would just be to package Qt and disable SDL. It's missing a bunch of configuration options and is going to be dropped whenever I finish merging it with nogui anyway, which won't rely on an external file selector.

PR #1348, opened 2020-12-30
Added compiled-in trace to file functionality

PugsyMAME:
Added tracer to cpu debugger menu


PugsyMAME:
Concerns expressed over speed issue.

PR #1354, opened 2020-12-30
Enabled Tracer functionality into cpu debugger menu

PugsyMAME:
Added standard slow debugger for users to use


stenzek:
Manually merged into master with corrections.

PR #1372, opened 2021-01-03
AnalogController: Refactor data transfer implementation

ggrtk:
`m_state` serialization is replaced with `m_command`, but this relies on save states happening outside of transfer sequences. Based on that assumption, transfer and receive buffers aren't being serialized in these changes.

### DIFF
--- a/src/core/analog_controller.cpp
+++ b/src/core/analog_controller.cpp
@@ -23,10 +23,12 @@ ControllerType AnalogController::GetType() const
 
 void AnalogController::Reset()
 {
-  m_state = State::Idle;
+  m_command = Command::Idle;
+  m_command_step = 0;
+  m_rx_buffer.fill(0x00);
+  m_tx_buffer.fill(0x00);
   m_analog_mode = false;
   m_configuration_mode = false;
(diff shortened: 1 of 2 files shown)


src/core/analog_controller.cpp
```diff
@@ -271,350 +262,370 @@ void AnalogController::SetMotorStateForConfigIndex(int index, u8 value)
...
-      // e.g. need all 0xFFs?
+  if (m_command_step == 0)
+  {
+    m_command = Command::Idle;
 
-      m_state = State::Idle;
-      ack = false;
-    }
-    break;
+    std::string tmp_str = "";
```
stenzek, line 616: Not really a fan of the heap allocations here - can we put this behind a `#ifdef _DEBUG`? (or use `SmallString`, which is stack-allocated).

But otherwise a nice cleanup!

ggrtk: Ah shoot -- meant to get rid of the heap allocation (since it's clear anyways what the transfer lengths are), but forgot to do so. Thanks for the catch!

PR #1470, opened 2021-01-16
Update PL translation

CookiePLMonster:
A big update to the Polish localization, correcting strings as well as translating some missing ones. As of the time of submitting this PR, **all** strings are translated and no obsolete string is present in the `.ts` file.

Additionally, this PR modifies the Game List widget to allow for proper button sizing with multi-line strings.


Bekon241:
If i can help i do that :)

PR #1515, opened 2021-01-26
UI: Add missing register widget help

Max833:
I think "Enable TTY Output" is the last missing widget help text. Unfortunately, I have no idea what it does. ^^;

Thanks for your awesome emulator, stenzek! :)


stenzek:
Thanks! For what it's worth, tty logging is basically printf() for the games, some have debug messages. But because the printf executes out of the BIOS ROM, it's very slow, and can cause games to crash.

PR #1528, opened 2021-01-28
Android Updated Some Missing Arrays

andercard0:
Updated some missing arrays for Portuguese language translation.


stenzek:
Fixed and committed in https://github.com/stenzek/duckstation/commit/b43b24ce8d2288ac97c54f655f105a687ec9b21c

PR #1543, opened 2021-01-31
Don't try to compile imgui code unless WITH_IMGUI is defined.

MaddTheSane:


stenzek:
I deliberately removed these because there's no build environment where imgui isn't built, and having ifdefs all over the place just makes things messy, especially when the functions are defined anyway.

MaddTheSane:
All I knew was that my OpenEmu plug-in wasn't compiling anymore. However, if imgui is now planned to always be built, I'll just build and include imgui with the plug-in.

stenzek:
Probably better that way. It shouldn't add much to the code size anyway.

The host display classes in frontend common are mainly intended for frontend not core/plugin use, there'd be a fair bit off dead code I'd think. May be better off writing your own minimal hostdisplay implementation.

Issue #1561, opened 2021-02-01
Eurocase EUGA controller not recognized


MortimerScharf:
**Description of the issue / bug**
My EUGA Eurocase controller (a third-party controller) is not being recognized by this emulator. All the other emulators recognize it no problem. It is recognized as "PLAYSTATION(R)3CONTEROLLER" (sic) by Windows, 

**Controller Troubleshoot Report**
1. Have you installed any drivers or wrappers on your system, or do you have any programs like Steam open? The controller is plug and play, needs no drivers. And yes, I do have Steam open, but closing it does not seem to affect it.
2. Which controller backend are you using in general settings? I tried all of them, none of them work except for SDL which only recognizes one or two buttons from the gamepad itself, ignoring all other inputs.

**Steps to reproduce**
I simply connected my gamepad, tried to bind the keys as I usually do with other emulators but this emulator did not recognize my gamepad at all. I had to use joy2key to use this EUROCASE gamepad.

**Enhancements information**
They are turned off.

**Hardware/software**
Windows 7 Ultimate, Intel 7700, GeForce GTX 1050. OpenGL.

**Emulator version**
0.1-3109-gfd269ec9

ggrtk:
Follow the instructions in the SDL_GameControllerDB repository that is linked in this section of the README https://github.com/stenzek/duckstation#sdl-game-controller-database for creating a new mapping, and add the new mapping to the `gamecontrollerdb.txt` file in your DuckStation user directory.

MortimerScharf:
I tried to do that first before posting this topic but I could not really understand what I needed to do. I noticed that part in the README file but I could not really understand what to do.

stenzek:
Try using DInput directly instead.

PR #1569, opened 2021-02-03
README.md: Update SDL Game Controller Database section

ggrtk:
README clarification for #1561.

On a side note, might be worth moving the build instructions to a BUILDING.md to reduce clutter, since I'd expect anyone building from source to be able to navigate a repository.


stenzek:
> On a side note, might be worth moving the build instructions to a BUILDING.md to reduce clutter, since I'd expect anyone building from source to be able to navigate a repository.

That's a good idea, thanks for the suggestion.

PR #1580, opened 2021-02-04
Update shadergen.cpp

MaddTheSane:
Quiets a warning in Clang about copying variables.

Actual warning text: 
> Loop variable '[qualifiers, name]' of type 'const std::__1::pair<const char *, const char *>' creates a copy from type 'const std::__1::pair<const char *, const char *>'
> Use reference type 'const std::__1::pair<const char *, const char *> &' to prevent copying


stenzek:
Seems like a silly warning in this case - it's just pointer types, unpacking them should actually be efficient than the extra level of indirection from accessing through through a reference.

PR #1581, opened 2021-02-05
Add printf-like macro to Writef.

MaddTheSane:
This should help catch formatting bugs.


MaddTheSane:
Looks like the build failures on Android are due to a couple of places using `Log_ErrorPrintf(message)` instead of `Log_ErrorPrint(message)` or `Log_ErrorPrintf("%s", message)` due to `-Werror,-Wformat-security` being set (either as a default or at compile time).

PR #1595, opened 2021-02-07
Update Strings Pt-Br

andercard0:
Updated to latest. 

Note: 

- OSD still missing

- There is a "r" letter that i guess need to be removed as follows: 

duckstation / android /app/ src/ main/ res/ values-pt-rBR

Should be: pt-BR only


stenzek:
> There is a "r" letter that i guess need to be removed as follows:

Are you sure about this? Android Studio created the directory, and it loads fine.

PR #1622, opened 2021-02-09
Add Y values for aspect scaling

Muzzarino:


stenzek:
Not really sure what you're trying to accomplish here?

Muzzarino:
> Not really sure what you're trying to accomplish here?

The vertical fov should be scaled if below the native aspect ratio rather than cropping off the horizontal from 4:3.

stenzek:
Unless I'm reading this wrong, you're effectively setting the screen-space coordinates to 1 for PGXP (which will break things), and using X (IR1) for both X and Y for native coordinates, and the `* 1 / 1` is redundant - so I don't see how this will produce correct results.

Muzzarino:
> Unless I'm reading this wrong, you're effectively setting the screen-space coordinates to 1 for PGXP (which will break things), and using X (IR1) for both X and Y for native coordinates, and the `* 1 / 1` is redundant - so I don't see how this will produce correct results.



> Unless I'm reading this wrong, you're effectively setting the screen-space coordinates to 1 for PGXP (which will break things), and using X (IR1) for both X and Y for native coordinates, and the `* 1 / 1` is redundant - so I don't see how this will produce correct results.

It's to maintain the fov at it's original scale while the other is being extended.

stenzek:
Have you actually tested this patch? I can't see how it would work at all with the current diff.

stenzek:
Closing since the changes here break rendering. Happy to reconsider/merge if you fix it.

PR #1646, opened 2021-02-16
optimize locking mechanism for OSD messages (deadlock-proofing)

jstine35:
Added benefits:
 - uses newly-added EnumerateOSDMessages for the Common UI OSD, same as is used for Fullscreen UI.

It's always a good idea to implement message queues in the fashion I've done here, to avoid the edge case where a callback GUI function might do something that itself blocks against some other mutex and then poses a risk for deadlock. Since the emulator doesn't really make use of threads much at this point, there's not a lot of immediate tangible benefit to this change. It's really more of an academic change, with potentially some future-proofing qualities.

I also removed some dead code from FullscreenUI. The enumerator handles the `display_show_osd_messages` setting internally now so FullscreenUI didn't need a special case handler anymore.

### DIFF
--- a/src/frontend-common/common_host_interface.cpp
+++ b/src/frontend-common/common_host_interface.cpp
@@ -930,30 +930,26 @@ void CommonHostInterface::AddOSDMessage(std::string message, float duration /*= 
   msg.duration = duration;
 
   std::unique_lock<std::mutex> lock(m_osd_messages_lock);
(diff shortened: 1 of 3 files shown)


stenzek:
Seems reasonable, although there isn't really multiple threads writing to the OSD here in the first place (except maybe occasionally in Qt).

PR #1657, opened 2021-02-17
Save State UI: Add legend showing hotkeys

CookiePLMonster:
This PR adds a legend to the Save State Selector UI. In my experience, I kept misremembering Save/Load buttons and using the wrong one when on this screen, so a legend should help.

To do:
* ~~Style the legend better? Most notably, some margins may be needed.~~
* ~~Strip the device name maybe - `F1` or `Button14` might be good enough in this case.~~
* ~~Make the legend translatable~~

Current look:
image
image

### DIFF
--- a/src/frontend-common/save_state_selector_ui.cpp
+++ b/src/frontend-common/save_state_selector_ui.cpp
@@ -187,52 +187,105 @@ void SaveStateSelectorUI::Draw()
   ImGui::PushStyleVar(ImGuiStyleVar_WindowRounding, rounding);
 
   if (ImGui::Begin("##save_state_selector", nullptr,
-                   ImGuiWindowFlags_NoFocusOnAppearing | ImGuiWindowFlags_NoInputs | ImGuiWindowFlags_NoTitleBar))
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
Changed the following:
* Added a horizontal margin to the legend. I tried adding a vertical margin and it bloated the panel more than needed IMO.
* Stripped the device names from buttons.
* Made the entire widget translatable. I included a showcase using the Polish translation, but it's not part of this PR.

src/frontend-common/save_state_selector_ui.cpp
```diff
@@ -187,52 +187,120 @@ void SaveStateSelectorUI::Draw()
...
+          result = str.substr(slash_pos + 1);
+        }
+        else
+        {
+          result = std::move(str);
+        }
+        return result;
+      };
+
+      const std::string load_savestate_hotkey =
```
stenzek, line 274: Not really a fan of having these lookups - I try to avoid per-frame heap allocations. Could we cache/stash them somewhere in the class?

CookiePLMonster: Sure, but then we need to subscribe to a signal if those ever change, I think :(

PR #1659, opened 2021-02-18
pad: improve messaging when UI config doesn't match savestate config, and document some edge cases

jstine35:
- Adds 8x controllers and memcards into the savestate.
 - maintains backward compatibility
 - simplifies some aspects of the logic used to load states
 - renames NUM_SLOTS to NUM_DEVICES because slots are ambiguous (2 sio slots on a console, 4 slots per mtap, sometimes one or both are called 'ports' instead, etc)

This PR can be merged as it is, or choose to wait until the full multitap impl on the emu backend is finished (eg, Incremental strategy for better reviews vs. a monolithic commit that bumps the savestate version # only once). I'm expecting to have the multitap functionality part ready before 2/22.

### testing

I ran lots of tests loading states from prev version that contained about 6x different variations of controller and memcard bindings on the host UI.

### for discussion

One of the typical caveats of emulating mtap is mapping the controllers. Not all games enumerate mtap controllers in the same way. For example, if a user has an mtap on slot 1 and a regular controller on slot 2 of the console:

device 1 -> mtap 1 slot 1
device 2 -> console slot 2
device 3 -> mtap 1 slot 2

... or ...

device 1 -> mtap 1 slot 1
device 2 -> mtap1 slot 2
device 2 -> console slot 2

This shouldn't impact how savestates are handled in any case. It should probably be implemented as a game-specific remap setting that just changes how emu devices 1->8 are shown to the user in the UI.

### DIFF
--- a/src/core/pad.cpp
+++ b/src/core/pad.cpp
@@ -67,40 +67,19 @@ bool Pad::DoState(StateWrapper& sw)
     {
       if (controller_type != state_controller_type)
       {
-        if (g_settings.load_devices_from_save_states)
-        {
-          g_host_interface->AddFormattedOSDMessage(
-            10.0f,
-            g_host_interface->TranslateString(
-              "OSDMessage", "Save state contains controller type %s in port %u, but %s is used. Switching."),
-            Settings::GetControllerTypeName(state_controller_type), i + 1u,
-            Settings::GetControllerTypeName(controller_type));
+        g_host_interface->AddFormattedOSDMessage(
+          10.0f,
+          g_host_interface->TranslateString(
+            "OSDMessage", "Save state contains controller type %s in port %u, but %s is used. Switching."),
+          Settings::GetControllerTypeName(state_controller_type), i + 1u,
+          Settings::GetControllerTypeName(controller_type));
 
-          m_controllers[i].reset();
-          if (state_controller_type != ControllerType::None)
-          {
-            m_controllers[i] = Controller::Create(state_controller_type, i);
-            if (!sw.DoMarker("Controller") || !m_controllers[i]->DoState(sw, false))
-              return false;
-          }
-        }
-        else
+        m_controllers[i].reset();
+        if (state_controller_type != ControllerType::None)
         {
-          g_host_interface->AddFormattedOSDMessage(
-            10.0f,
-            g_host_interface->TranslateString("OSDMessage", "Ignoring mismatched controller type %s in port %u."),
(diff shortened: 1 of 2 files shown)


ggrtk:
I have multitap functionality and the UI update mostly finished on a branch already, just haven't gotten around to running hardware tests to determine some edge cases yet. Need to check the access timings as well, but it should be safe to just use the current ones and ignore any buffering/routing in the multitap itself.

I can get around to wrapping it up sometime this weekend or next week. Maybe you can rebase anything that's needed from this PR on top of it?

jstine35:
Yeah, I can rebase later.

As for timing - The multitap was designed to not require buffering. The multitap can (and should) be implemented without any buffer within the emulator. It requires only two state variables. If there's a buffer involved, then it can be simplified - and an implementation that does not have a buffer will certainly more closely resemble the original.

This doesn't rule out the possibility that the multitap may have some peculiar extra delays while writing out its results from the connected pad(s). But that would be fairly pedantic. The psx pad library was fairly tolerant of timing on pad reads by design, so long as the IRQ has some initial delay (typical of that era, they would send a command and then set up some IRQ handling immediately after, so a few cycles delay is always a good idea, whether it's SIO, DMA, GPU, etc)

stenzek:
Many thanks! I'll give this a proper review soon.

Was going to point out that you can avoid breaking old states by using the version number, but from a quick glance it looks like you've figured that already! :)

ggrtk:
It's going to be a while longer from my end. Compatibility is pretty high now, but a couple games are relying on undocumented hardware behavior that I'll need to reverse.

legend-of-life:
Crash bash and crash team racing is support 4 player and I played by my friends in epsxe emulator but in duckstation I can't because that support 2 game pad, please add 4 gamepad support to duckstation.

jstine35:
Updated this on top of @ggrtk's multitap. Renamed the PR since the scope is now reduced:

The change cleans up and documents some of the various edge cases relating to UI messaging and behavior when the user's config and the savestate pad mapping do not match. 

The refactoring goal was to more clearly define the differences in behavior between Reading and Writing modes, and to help ensure that correct behavior is maintained -- specifically that if controllers/memcards are present in the savestate that they _for sure_ always get loaded out of the state _even if the UI forcibly ignores them_ (in order to advance savestate read pointers properly).

stenzek:
Sound good, thanks! Will take a look soon.

PR #1664, opened 2021-02-21
Update Italian Translation (Android)

RaydenX93:
Let's see if the CI likes it so far...


RaydenX93:
@stenzek I guess I have to escape the characters then?

RaydenX93:
It's not going to happen anytime soon.
Better to start from scratch.

PR #1674, opened 2021-02-23
Fix for batch mode nogui

crb04c:


stenzek:
Ah, left over from when NoGUI didn't have any UI at all...

PR #1677, opened 2021-02-23
Implement RetroAchivements

stenzek:
Opening as a PR since people want to test it.

Still remaining TODO:

- [ ] Linux support (only has a Windows HTTP downloader for now)
- [ ] Better handling of login errors (don't block the UI)
- [ ] Login UI for big duck mode
- [ ] Handle re-login when token is invalid
- [ ] Implement hardcode mode and restrictions
- [ ] Android UI

What works:

- Login, unlock list downloads, achievement granting
- Popups and achievement browser
- Settings in Qt frontend

**NOTE: FULLSCREEN UI MUST BE ENABLED IN GENERAL SETTINGS TO SEE ACHIEVEMENT POPUPS**


andercard0:
Feedback and Suggestions:

- While Open Quick Menu or even in Big Duck Mode does not pause the game
(I'd suggest a option as we have "pause on focus loss")

- Shortcut exit full-screen is not working in Big Duck Mode
tried to change / bind another key for "toggle full-screen button" it prompt me
to hit other key I choose other key on the keyboard but duck does not bind it at all.

Even, if my shortcut (F11) key is assigned already it does not work. 
Tried to edit (settings.ini) removing F11 for full-screen and no deal; i got into 
shortcut screen, bind shortcut, timer goes 5..4..3..2.. but any key isn't being accepted.

- Would be great if we have Cheevos retroactively. meaning, as soon as my MC is loaded it 
'see' my progress there and this way upgrade my Cheevos progress as well.

Really do not  know  if this is viable but would be a good thing
so we would not have to replay our games again :p

- Source text in English .ts is 'Gane Slot' instead Game Slot.

stenzek:
> * While Open Quick Menu or even in Big Duck Mode does not pause the game (I'd suggest a option as we have "pause on focus loss")

Added "Pause On Menu" option.

> * Would be great if we have Cheevos retroactively. meaning, as soon as my MC is loaded it 'see' my progress there and this way upgrade my Cheevos progress as well.

I suspect not, unless the memory the achievement is looking for happens to have the correct value.

> * Source text in English .ts is 'Gane Slot' instead Game Slot.

Can't see this anywhere?

> * when login fails duckstation does not give me another chance to redo / type my credentials (as you may view below) i have to hit close window and retry.

Fixed.

> * Cheevos aren't being updated at Retro Achievements site and / or when updates it, information of the current dates aren't accurate i.e down below

Fixed.

> * Don't know if this is important but as you may view duck is not being showed. Checked resources folder and duck.png is there.

Might be a bad logo.png replacement. Remove it and see if that changes anything.

> * When in big duck mode while fast forward FF volume is too loud as you can see both are set up at 16 of volume but when i Fast forward the game goes pretty loud. instead of being low

Fixed.

andercard0:
Just tried to use No GUI with your new commits but it does not run it starts to load the nothing happen.

Only QT is working
image

andercard0:
Feedback 2

- Only QT is working no-GUI still loads then fails;

- I open quick menu and emulation pauses but yet have to use my mouse to choose resume game to close it property and also have to unpause the emulation via Pause button in toolbar. 

Note: this happens only while in Full-screen mode.

If possible ideal scenario would be a hotkey in order to open quick menu as well as to close it. likewise i do with full-screen on and off with F11 for example.

- Other thing I've noticed is as soon as i open quick menu there is no way (as said earlier) to return with controls but, if i use LB or RB the options are highlighted but even so i can't choose any option with my controller.

andercard0:
Feedback 3

Feedback 3

- QT - Controllers aren't being used correctly
Made a video in order to explain this issue better take a look where you can please
At 00:19 when Resume Game and Exit Game are highlighted it was me pressing L1 / R1
this is the only input i can do in that screen; While in it game is paused. 

When choose Achievements options game is unpased, as already reported earlier, so beyond
00:26 game is running which Cheevos screen and all that but when in Fullscreen it's a bit
annoying at this point, cause my mouse pointer isn't visible.

Video: https://streamable.com/p0w7ly

If you allow me a suggestion (for QT) would be, when quick menu open game keep running but
in a low volume and 'turn off' game input; if you notice when Cheevos screen is opened
if you press any button game is responding to it instead let me navigate trough cheeevos 
list. I only can do it via mouse and keyboard arrow up and dow which again, i set up those 
for vol increase (UP) and vol decrease (down) both is being accepted.

- Returning to game from: cheevos, load, save, cheats menus are perfect!

- Not sure if this is suppose to happen but after click on "Change Disc" > Close (X)
it return to the game but in a paused state.

Note: you'll not see my mouse pointer at the end but believe me had to click pause button.

Video: https://streamable.com/d2h2zm

- Crash while opening Cheevos Settings
Was testing Settings > Achievements Settings Emulator hangs and crash. 

- Big Duck Mode (NoGui)

- Crash while opening Cheevos Settings Same behavior as in QT
Was testing Settings > Achievements Settings Emulator hangs and crash.

- DuckStation Minimizes as soon as i click / change focus to another monitor

- While in Cheevos list menu is not possible to close it with my controller had to X it 
to return to game play and it return game is paused even choosing quick menu, resume, nothing
happens had to restart the game. After turn new option "pause on menu" this did not happen.

stenzek:
Crash in big duck mode should be fixed. Probably should do some quick re-testing as I had to switch from wininet to winhttp to support POST requests (for rich presence).

andercard0:
> Crash in big duck mode should be fixed. Probably should do some quick re-testing as I had to switch from wininet to winhttp to support POST requests (for rich presence).

Yup! fixed indeed. only the other reported issues still remaining.

PR #1678, opened 2021-02-23
Filter out Keypad as a mod in sdl_key_names

crb04c:


stenzek:
Nice catch. Made it a little safer, not that there's any other modifiers that will come up.

PR #1679, opened 2021-02-23
Allow keyboard binding in nogui/fullscreen mode

crb04c:
Couldn't think of a better way to abstract this currently


stenzek:
Thanks! I was deferring writing this until I refactor the input binding system to support chords (multi-button combinations), but this is a good fix in the meantime.

PR #1687, opened 2021-02-25
FullscreenUI: Update type cache when controller type is changed

ggrtk:


src/frontend-common/fullscreen_ui.cpp
```diff
@@ -1540,7 +1540,7 @@ void DrawSettingsWindow()
           CommonHostInterface::InputProfileList profiles(s_host_interface->GetInputProfileList());
           ImGuiFullscreen::ChoiceDialogOptions options;
           options.reserve(profiles.size());
-          for (const CommonHostInterface::InputProfileEntry& entry : profiles)
+          for (CommonHostInterface::InputProfileEntry& entry : profiles)
```
CookiePLMonster, line 1543: Intended?

ggrtk: Intentional change, since with const we get copy constructor behavior rather than move constructor behavior. I think the latter was what was originally intended.

CookiePLMonster: Oh, yeah! I think that could be modified to an one-liner but I'm not sure how currently, and the current way seems fine.

stenzek: Yep, my bad here.

PR #1694, opened 2021-02-26
Settings: Always save controller type to settings interface

ggrtk:


stenzek:
I swear I changed this a while back... must've got lost on some branch somewhere :)

PR #1697, opened 2021-02-27
Core: Add Multitap support

ggrtk:


src/core/settings.cpp
```diff
@@ -77,6 +78,31 @@ bool Settings::HasAnyPerGameMemoryCards() const
...
 
+bool Settings::IsMultitapEnabledOnPort(u32 port) const
+{
+  if (port < NUM_MULTITAPS)
+  {
+    switch (multitap_mode)
+    {
+      case MultitapMode::Disabled:
+        return false;
+        break;
```
stenzek, line 89: Nit: `break` is redundant here

stenzek (commented):
Many thanks for all your hard work here!

We'll probably want to rebase #1659 off this for handing saving/loading, but for now skipping loading controllers for old states sounds reasonable to me.

Issue #1065, opened 2020-11-12
[Enhancement/Android Standalone] Improve holding buttons on touchscreens


PookaMustard:
**Is your feature request related to a problem? Please describe.**
The touchscreen used by Duckstation isn't quite up to par for stuff like jumping while charging (the Spyro series), or holding all four shoulder buttons at once, or any other combo in which you press a button while holding another.

**Describe the solution you'd like**
I propose that the first button you hold remains held as long as you don't release your finger, and any buttons you swipe over are pressed/held as well until you swipe away. For an example:

1. Hold square
2. Swipe away from square: square remains held
3. Swipe over to X/cross: it is now held
4. Swipe away from X/cross: it is now released
5. Take your finger off the screen: Square is now released.

This feature can be configurable such that it you can enable or disable it for the D-pad. A live example of this can be experienced in the [Super Mario 64 native mobile port](https://github.com/JustMeDaFaq/sm64-port-android). It can work with multiple taps as well, providing support for features such as holding all four shoulder buttons with just two fingers.

**Describe alternatives you've considered**
None so far.

**Additional context**
When playing the aforementioned Super Mario 64 mobile port, I found it extremely easy to do stuff like long jumps, thanks to the fact I could hold Z/Trigger and swipe over to the A button without Z/Trigger getting released. I can imagine if this little trick did wonders for SM64, it would for many games, even if it needs to be enabled via the settings screen.

clod2692:
Yes, buttons remap is absolutely necessary, i want to add that xbox elite controller's triggers don't work, but a remap option maybe could fix this issue

PookaMustard:
Closed with the merging of #1723.

PR #1723, opened 2021-03-02
Add touch gliding

PookaMustard:
This PR adds the functionality I requested in #1065, that is, touch gliding (for lack of a better name) for Android. The commit has been tested extensively for the better part of a day on a real phone.

This basically keeps the first button held as long as your finger keeps touching the screen. This lets you hold square and then glide/slide over to X to charge and jump as Spyro with minimal effort, using your touchscreen. It also supports multiple touch traces, perfect for games like, I don't know, Crash Team Racing (X to Circle for using items without losing speed, R1<>L1 for drifting and boosting, both can happen concurrently).

How does this work:
A map (mFirstHolds) is available for linking touch pointer IDs to button views. When you touch the screen, a pointer is made. Now, when your pointer starts on, or goes to the first button, the map saves the pointerID>buttonview combo. Any future buttons this pointer slides over are not saved in the combo, but will be activated, and when you slide away it will disable, but the first button remains on. You can do this for as many fingers as you can possibly get on the screen, and they'll be treated separately.

The moment one of your pointers let go of the screen, the combo that was mapped is deleted. When all pointers are released, the map is wholly cleared.

### DIFF
--- a/android/app/src/main/java/com/github/stenzek/duckstation/TouchscreenControllerView.java
+++ b/android/app/src/main/java/com/github/stenzek/duckstation/TouchscreenControllerView.java
@@ -20,6 +20,9 @@ import androidx.constraintlayout.widget.ConstraintLayout;
 import androidx.preference.PreferenceManager;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * TODO: document your custom view class.
@@ -42,6 +45,8 @@ public class TouchscreenControllerView extends FrameLayout {
     private float mMovingLastY = 0.0f;
     private ConstraintLayout mEditLayout = null;
     private int mOpacity = 100;
+    private Map<Integer, View> mFirstHolds = new HashMap<>();
+    private ArrayList<String> mDpadButtons = new ArrayList<String>(Arrays.asList("UpButton", "DownButton", "LeftButton", "RightButton"));
 
     public TouchscreenControllerView(Context context) {
         super(context);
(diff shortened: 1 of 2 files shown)


android/app/src/main/java/com/github/stenzek/duckstation/TouchscreenControllerView.java
```diff
@@ -42,6 +45,8 @@
     private float mMovingLastY = 0.0f;
     private ConstraintLayout mEditLayout = null;
     private int mOpacity = 100;
+    private Map<Integer, View> mFirstHolds = new HashMap<>();
+    private ArrayList<String> mDpadButtons = new ArrayList<String>(Arrays.asList("UpButton", "DownButton", "LeftButton", "RightButton"));
```
stenzek, line 49: Probably better to make this an attribute of the button itself, if for some reason we wanted to extend it to non-dpad buttons in the future.

stenzek:
Do you think it's worth having this as a toggle-able option? Or would always-on be okay.

android/app/src/main/java/com/github/stenzek/duckstation/TouchscreenControllerView.java
PookaMustard: That sounds like a great idea actually. Better than what I have here in mind. I guess the buttonview object should have an attribute "is it glidable" during linkButton, as well as a command to fetch it later.

PookaMustard:
> Do you think it's worth having this as a toggle-able option? Or would always-on be okay.

I would totally make it an option if I knew how.

PookaMustard:
Okay then, I made the glidability an attribute. On testing, everything's OK!

PR #1724, opened 2021-03-02
Cheevos: Implement hardcore mode

stenzek:
PR'ing because I still need to check that everything which needs to be disabled, has been disabled.

TODO;

- [x] Implement toggle in fullscreen UI/big duck settings

### DIFF
--- a/src/frontend-common/fullscreen_ui.cpp
+++ b/src/frontend-common/fullscreen_ui.cpp
@@ -93,6 +93,15 @@ static void OpenAboutWindow();
 static void SetDebugMenuEnabled(bool enabled);
 static void UpdateDebugMenuVisibility();
 
+static ALWAYS_INLINE bool IsCheevosHardcoreModeActive()
+{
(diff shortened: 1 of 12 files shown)


src/frontend-common/fullscreen_ui.cpp
```diff
@@ -1066,6 +1075,91 @@ static bool ToggleButtonForNonSetting(const char* title, const char* summary, co
...
+  ImGui::PushFont(g_large_font);
+
+  bool is_open = true;
+  if (ImGui::BeginPopupModal("Achievements Login", &is_open, ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize))
+  {
+
+    ImGui::TextWrapped("Please enter user name and password for retroachievements.org.");
+    ImGui::NewLine();
+    ImGui::TextWrapped(
+      "Your password will not be saved in DuckStation, instead an access token will be generated and used instead.");
```
CookiePLMonster, line 1094: Double `instead`.

PR #1726, opened 2021-03-02
Misc Qt stuff

CookiePLMonster:
**IMPORTANT NOTE:** If it's not done automatically, `duckstation-qt_en.qm` must be added to the artifacts!

1. Hooks up plurality for English. So far very few strings utilize it, because we can't yet use plurality outside of Qt strings. It'll be much more useful once it's in there though, for strings like `X cheats loaded` etc.
2. "Stretch to fill" had a broken widget help assignment.
3. As far as I can tell, `QStringLiteral("%1").arg(num)` can be simplified to `QString::number(num)`, so I did a full sweep through the codebase.

### DIFF
--- a/src/duckstation-qt/qthostinterface.cpp
+++ b/src/duckstation-qt/qthostinterface.cpp
@@ -64,7 +64,7 @@ const char* QtHostInterface::GetFrontendName() const
 
 std::vector<std::pair<QString, QString>> QtHostInterface::getAvailableLanguageList()
 {
-  return {{QStringLiteral("English"), QStringLiteral("")},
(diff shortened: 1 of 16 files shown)


src/duckstation-qt/qthostinterface.cpp
```diff
@@ -129,7 +129,7 @@ void QtHostInterface::installTranslator()
 
   std::string language = GetStringSettingValue("Main", "Language", "");
```
stenzek, line 130: Last parameter here is the default, so this can be
```cpp
const std::string language(GetStringSettingValue("Main", "Language", "en"));
```

Issue #1722, opened 2021-03-02
DuckStation needs proper Cheat Manager Support


titankilla:
Hello There.

Stenzek we spoke already, and here is the request as you asked.

Could you please find time at some point to add more features to Cheat Manager ? yes i dont like cheats alot myself, but as i exlained to you i use "visual" cheats mostly and there needs to be some changes in the code to add "Pop Up" window and have working cheat tables...

i dont know how many other PS 1 games (besides Front Mission 3) require this "pop up window- and cheat tables) but i would think there are some - and without it (ppl new to this wont be able to manually use certain cheat codes - cuz they dont know what to do...)

"PEC" "psx emulation cheater" - was programmed to my understanding as following:
if there is a table code at the end "%BP" it would do a pop up window and allow you to chose cheat codes from the table to replace the question marks at the end of code 3011ABA3 00?? - with values from that table ...
3011ABA3 00??
%BP
it's very easy to understand how PEC works and add your own cheats and tables (which i did - i have fixed bugged default FM 3 codes and added new ones and alot of custom tables)  - and would be cool if DuckStation had this exact feature ... you could download PEC urself and just check how it's working and maybe you can make ur own solution or make it even better 

Thank you and have a nice day.

![2352523535](https://user-images.githubusercontent.com/64417061/109618025-5f773e00-7b37-11eb-976d-610c27897030.jpg)

PR #1736, opened 2021-03-03
Added new cheat cheat types: C3, C4, C5, C6, D7&52

PugsyMAME:
C3-C7 are variants of C0
D7 is a BIT based joker to rule them all. It includes the analog sticks (@ggrtk thanks for adding the analog reading capability). Also added the facility of making dual single key joker by varying the amount of time a button is held down.
51 is a complicated beast that I still need to document

* C3XXXXXX 00YY - 8-Bit Master Code, if ($XXXXXX) is less than 0xYY poke all
  00000000 FFFF       following codes for rest of the cheat or until it reaches the
                      00000000 FFFF line.
* C4XXXXXX 00YY - 8-Bit Master Code, if ($XXXXXX) is greater than 0xYY poke all
  00000000 FFFF       following codes for rest of the cheat or until it reaches the
                      00000000 FFFF line.
* C5XXXXXX YYYY - 16-Bit Master Code, if ($XXXXXX) is less than 0xYYYY poke all
  00000000 FFFF       following codes for rest of the cheat or until it reaches the
                      00000000 FFFF line.
* C6XXXXXX YYYY - 16-Bit Master Code, if ($XXXXXX) is greater than 0xYYYY poke all
  00000000 FFFF       following codes for rest of the cheat or until it reaches the
                      00000000 FFFF line.
* D7PQRRRR TTYYYYYY - 24-Bit Universal BIT Joker, OR the hex values to
                     combine into a multi-button joker. Because it is BIT
                     based it is better than D4, D5, D6 or using a D0 joker as you
                     do not need to worry about any other buttons being
                     pressed at the same time and you get both analog
                     sticks for extra functionality. Note if you want to use it
                     just as a enhanced joker just use D7000000 00YYYYYY when
                     the buttons/directions are pressed or D7100000 00YYYYYY
                     when you want to ensure they are not all pressed.
                     QRRRR TT provides the capability of only activating the
                     following codes after the keys have been held in for a set
                     amount of frames. 003C = 60 Frames = 1 Second at 100% Speed
                        YYYYYY = 000001 L2 Button
                        YYYYYY = 000002 R2 Button
                        YYYYYY = 000004 L1 Button
                        YYYYYY = 000008 R1 Button
                        YYYYYY = 000010 Triangle Button
                        YYYYYY = 000020 Circle Button
                        YYYYYY = 000040 X Button
                        YYYYYY = 000080 Square Button
                        YYYYYY = 000100 Select Button
                        YYYYYY = 000200 L3 Button
                        YYYYYY = 000400 R3 Button
                        YYYYYY = 000800 Start Button
                        YYYYYY = 001000 Up (Digital)
                        YYYYYY = 002000 Right (Digital)
                        YYYYYY = 004000 Down (Digital)
                        YYYYYY = 008000 Left (Digital)
                        YYYYYY = 010000 Up (Right Thumb)
                        YYYYYY = 020000 Right (Right Thumb)
                        YYYYYY = 040000 Down (Right Thumb)
                        YYYYYY = 080000 Left (Right Thumb)
                        YYYYYY = 100000 Up (Left Thumb)
                        YYYYYY = 200000 Right (Left Thumb)
                        YYYYYY = 400000 Down (Left Thumb)
                        YYYYYY = 800000 Left (Left Thumb)
                      NOTE: The 0s in the code are reserved for possible
                            future use.
                       TT=Temp Internal Variable 00-FF, 00 will mean it wont be
                            used, if it's not 00 do not use the same value for
                            jokers using different keypress combinations for
                            the same game.
                       P = 0 or 1. 0 = Check ALL YYYYYY Bits are ON
                                   1 = Check ALL YYYYYY Bits are OFF
                       Q = Frame Comparison 0 = Dont do any comparison
                                            1 = Check that the button combination
                                                has been held down for exactly
                                                RRRR frames.
                                            2 = Check that the button combination
                                                has been held down for more than
                                                RRRR frames.
                                            3 = Check that the button combination
                                                has been held down for less than
                                                RRRR frames.
                                            4 = Check that the button combination
                                                has been held down for anything
                                                but RRRR frames.
                       RRRR = 0000 to FFFF, Frame Comparison Value
                      It will then poke all following codes for rest of cheat
  00000000 FFFF       or until it reaches the 00000000 FFFF line.

### DIFF
--- a/src/core/cheats.cpp
+++ b/src/core/cheats.cpp
@@ -15,6 +15,7 @@
 #include <iomanip>
 #include <sstream>
 Log_SetChannel(Cheats);
+static std::array<u32, 256> temp_variable; //Used for D7 ,51 & 52 cheat types
 
 using KeyValuePairVector = std::vector<std::pair<std::string, std::string>>;
 
@@ -136,6 +137,58 @@ static u32 GetControllerButtonBits()
   }
 
   return translated_bits;
+}
+
+static u32 GetControllerAnalogBits()
+{
+  // 0x010000 - Right Thumb Up
+  // 0x020000 - Right Thumb Right
+  // 0x040000 - Right Thumb Down
+  // 0x080000 - Right Thumb Left
+  // 0x100000 - Left Thumb Up
+  // 0x200000 - Left Thumb Right
+  // 0x400000 - Left Thumb Down
+  // 0x800000 - Left Thumb Left
+
+  u32 bits = 0;
+  u8 l_ypos = 0;
+  u8 l_xpos = 0;
+  u8 r_ypos = 0;
+  u8 r_xpos = 0;
+
+  std::optional<u32> analog = 0;
+  for (u32 i = 0; i < NUM_CONTROLLER_AND_CARD_PORTS; i++)
+  {
+    Controller* controller = System::GetController(i);
+    if (!controller)
+      continue;
+
+    analog = controller->GetAnalogInputBytes();
+    if (analog.has_value())
+    {
+      l_ypos = Truncate8((analog.value() & 0xFF000000u) >> 24);
+      l_xpos = Truncate8((analog.value() & 0x00FF0000u) >> 16);
+      r_ypos = Truncate8((analog.value() & 0x0000FF00u) >> 8);
+      r_xpos = Truncate8(analog.value() & 0x000000FFu);
+      if (l_ypos < 0x50)
+        bits |= 0x100000;
+      else if (l_ypos > 0xA0)
+        bits |= 0x400000;
+      if (l_xpos < 0x50)
+        bits |= 0x800000;
+      else if (l_xpos > 0xA0)
+        bits |= 0x200000;
+      if (r_ypos < 0x50)
+        bits |= 0x10000;
+      else if (r_ypos > 0xA0)
+        bits |= 0x40000;
+      if (r_xpos < 0x50)
+        bits |= 0x80000;
+      else if (r_xpos > 0xA0)
+        bits |= 0x20000;
+    }
+  }
+  return bits;
 }
 
 CheatList::CheatList() = default;
@@ -629,11 +682,8 @@ bool CheatList::LoadFromString(const std::string& str, Format format)
     return LoadFromPCSXRString(str);
   else if (format == Format::Libretro)
     return LoadFromLibretroString(str);
-  else if (format == Format::EPSXe)
-    return LoadFromEPSXeString(str);
-
-  Log_ErrorPrintf("Invalid or unknown cheat format");
-  return false;
+  format = Format::EPSXe;
+  return LoadFromEPSXeString(str);
 }
 
 bool CheatList::SaveToPCSXRFile(const char* filename)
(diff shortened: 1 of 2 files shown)


PugsyMAME:
It's cheat type 51 rather than type 52...typo in commit message. oops

src/core/cheats.cpp
```diff
@@ -15,6 +15,7 @@
 #include <iomanip>
 #include <sstream>
 Log_SetChannel(Cheats);
+static std::array<u32, 256> temp_variable; //Used for D7 ,51 & 52 cheat types
```
stenzek, line 18: Do we want to make this storage per-code? Or share it between all codes.

Could be worth allocating it dynamically on demand if so.

PugsyMAME:
Thinking about it I got the nomenclature wrong, it's a register rather than a variable (it's probably because variable is how MAME refers to it)

Originally when I came up with the usage in the D7 type I wanted a register that was local to the cheat, but I didn't now how implement it realiablly with multiple D7s in a single cheat. So I put a global in as I knew that would definitely work.

However, I was asked about this feature request regarding Front Mission 3 cheats:-

https://github.com/stenzek/duckstation/issues/1722

This is partially the reason I came up with the 51 type, and using global registers makes it possible.

In the example above, one cheat would select %BP by poking a register

And other cheats would take the register contents and poke it into the picture modifier address for that character.

It would be ever be possible to use another register to store the addresses for the picture modifiers for every character.

You then would just select the character and enable a single picture modifer code that would take the picture modifier address from one register, and the poke it with the %BP value from another register.

Then you could have 8000 cheats represented with less than 400 entries.

MAME cheats use temp variables but they are locals, and for the vast majority of systems this is ideal but there have been times that I would have loved some globals.

The one downsize of global registers is that the cheat maker needs to be sure that any they use are unique for that game. But that checking for uniqueness only comes into play when a game has a existing non-related cheats using registers - which will be exceptionally rare.

stenzek:
Hmm. If the goal is to make a selectable option, wouldn't a dropdown in the UI be better?

But if there's uses outside of just selecting values for other codes, fair enough.

PugsyMAME:
Yes, a dropdown option would be MUCH better - but having looked at the problem with Front Mission 3 cheats that would be too far outside my comfort zone and a lot of work for something I'm not overly bothered about. I wouldn't know where to start to squeeze that PEC cheater functionality into the u32 u32 format without a major cheat engine rewrite into a scripting cheat engine, which at the current time would not be a good use of anyone's time. 

My solution at the very best is a workaround for the Front Mission 3 cheats. 

Type 51/52 is a method of squeezing some extra functionality out of the u32 u32 cheat format, I call it my Field of Dreams cheat type....build and they will come(maybe..).  In other words it's a solution waiting for the problems....

stenzek:
Okay, so if we add it as-is, do we want the 256-value array to be:
 - shared between codes, or per-code?
 - shared across frames, or reset every frame?

PugsyMAME:
Yes as is ideally. So shared across codes and across frames.

Just noticed, I included the temp fix so that Duckstation format cheats can be imported that I mentioned before. It's not an ideal fix, but in the short term its probably better than leaving it as is.

It replaces:
  else if (format == Format::EPSXe)
    return LoadFromEPSXeString(str);

  Log_ErrorPrintf("Invalid or unknown cheat format");
  return false;

With:
  format = Format::EPSXe;
  return LoadFromEPSXeString(str);

stenzek:
> Just noticed, I included the temp fix so that Duckstation format cheats can be imported that I mentioned before. It's not an ideal fix, but in the short term its probably better than leaving it as is.

Not really a fan of this, but I can't think of a better solution for now.

PR #1737, opened 2021-03-03
Update gamesettings.ini

6lackmag3:
force digital for Mortal Kombat 4

### DIFF
--- a/data/database/gamesettings.ini
+++ b/data/database/gamesettings.ini
@@ -828,3 +828,13 @@ DisableAnalogModeForcing = true
 # SLPS-00702 (Final Fantasy VII (Japan) (Disc 3))
 [SLPS-00702]
 DisableAnalogModeForcing = true
+
+
+# Mortal Kombat 4 (USA) (SLUS-00605)
+[SLUS-00605]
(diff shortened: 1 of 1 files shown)


ggrtk:
As a postmortem for this, turns out the game doesn't do anything special, it just has shoddy pad handling. The `DUAL SHOCK ENABLED/DISABLED` options in the menu is poorly named and just tells the software to attempt a rumble configuration rewrite and begin sending appropriate rumble bytes to any controllers that it had previously detected to be a DualShock controller (regardless of whether or not the rumble configuration succeeded, and even when the user has already replugged a different controller type).

6lackmag3:
@ggrtk  ok but it wont read inputs in Analog Mode anyway. Digital is indeed needed to play this game.

ggrtk:
Right, I was checking to rule out the possibility from the discussion yesterday with Stenzek that the game was relying on a DualShock toggling detection feature that we hadn't implemented. Hence my statement that the game's pad handling is just poorly programmed and isn't actually doing anything special.

stenzek:
Sounds good, many thanks for confirming @ggrtk !

PR #1762, opened 2021-03-08

zkdpower:

lack "mdf" in open menu, only "mds" current.

pressing hotkey to open "open quick menu" seems no use, and what is "quick menu",i never find it and can't see it in interface Menu. i want to test the "archivements" function.


stenzek:
> lack "mdf" in open menu, only "mds" current.

That's because MDFs can't be opened directly. You need the sidecar file (MDS).

> pressing hotkey to open "open quick menu" seems no use, and what is "quick menu",i never find it and can't see it in interface Menu. i want to test the "archivements" function.

You need to enable the fullscreen UI checkbox first. Or if you have achievements enabled, it will be automatically enabled.

PR #1769, opened 2021-03-08
Fix quick menu exit

CookiePLMonster:
Now requires the button to be released, **then** pressed. With this change, e0161c3bb2f5e5f4e9d7f9569ba86010899c08c5 can be reverted.

Should this logic be applied to `DrawChoiceDialog` and `DrawFileSelector` in `imgui_fullscreen.cpp`? If so, maybe it is better to move it there instead of `fullscreen_ui.cpp`.


src/frontend-common/fullscreen_ui.cpp
```diff
@@ -110,6 +109,7 @@ static bool s_debug_menu_allowed = false;
 static bool s_quick_menu_was_open = false;
 static bool s_was_paused_on_quick_menu_open = false;
 static bool s_about_window_open = false;
+static u32 s_close_button_state = 0;
```
stenzek, line 112: `s_close_button_pushed` would probably a better name, but *shrugs*.

PR #1792, opened 2021-03-12
Plurality for non-Qt messages

CookiePLMonster:
Extends `TranslateString` and `TranslateStdString` to support `disambiguation` and `n` parameters, and hooks them up to cheat OSD messages.

English and Polish translations are updated accordingly.

### DIFF
--- a/src/core/host_interface.cpp
+++ b/src/core/host_interface.cpp
@@ -992,14 +992,51 @@ float HostInterface::GetFloatSettingValue(const char* section, const char* key, 
   return float_value.value_or(default_value);
 }
 
-TinyString HostInterface::TranslateString(const char* context, const char* str) const
+TinyString HostInterface::TranslateString(const char* context, const char* str,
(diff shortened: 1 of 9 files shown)


src/core/host_interface.cpp
```diff
@@ -992,14 +992,51 @@ float HostInterface::GetFloatSettingValue(const char* section, const char* key,
...
-TinyString HostInterface::TranslateString(const char* context, const char* str) const
+TinyString HostInterface::TranslateString(const char* context, const char* str,
+                                          const char* disambiguation /*= nullptr*/, int n /*= -1*/) const
 {
-  return str;
+  TinyString result(str);
+  if (n >= 0)
+  {
+    const std::string number = std::to_string(n);
+    result.Replace("%n", number.c_str());
```
stenzek, line 1002: This could probably be a for loop over the string looking for `%`, but it's definitely not a hotspot at the moment.

PR #1820, opened 2021-03-15
Add skip presenting duplicate frames option

stenzek:
Let's see if this helps performance on low-end android devices/gpu readbacks...

### DIFF
--- a/src/core/host_display.cpp
+++ b/src/core/host_display.cpp
@@ -5,6 +5,7 @@
 #include "common/log.h"
 #include "common/string_util.h"
 #include "common/timer.h"
+#include "settings.h"
 #include "host_interface.h"
 #include "stb_image.h"
 #include "stb_image_resize.h"
@@ -27,6 +28,9 @@ void HostDisplay::SetDisplayMaxFPS(float max_fps)
 
 bool HostDisplay::ShouldSkipDisplayingFrame()
 {
+  if (!m_display_duplicate_frames && !m_display_changed)
+    return true;
+
   if (m_display_frame_interval == 0.0f)
     return false;
 
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -216,7 +216,6 @@ jobs:
         ./gradlew assembleRelease
 
     - name: Sign APK
-      if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/dev'
       uses: r0adkll/sign-android-release@v1
       with:
         releaseDirectory: android/app/build/outputs/apk/release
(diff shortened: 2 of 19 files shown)


andercard0:
Not sure if this suppose to affect my phone in any way but I'd like to report my experience and tests if you don't mind!

Quick testing it with: **Samsung J7 Prime** with CTR, Bust a Groove 2, Crash B. Warped and Gekido. only seen a significantly difference in crash **_WARPED_** which is running Okay-ish at 2x Res almost 20 ~ 28fps in Crash levels and in Coco's Bandicoot levels it does run at full speed in warp rooms and cut scenes there is no drops or buggy sounds at all.

Ofc. I did not beat the game to make sure all levels are okay played only 4 levels tops. and by the way, only have played with DS4 wired not in Bluetooth mode 'cause for me it does not work correctly personally have a huge lag when i try to use it, but i think this isn't emulator's fault. I'm just saying. 

**Gekido** runs at full speed with 2x of resolution too with no lags and issues too with this build/artifact provided 

**CRT** is a bit better but not so noticeable on my end tbh in menus and map screen full speed but as the race starts drops from: 30fps to 20 minimum and max 29fps when nothing but me is running meaning if I get the 1st place at the race and finally

**Bust a Groove 2** no deal, lags a lot when get into gameplay during cutscenes are okay 2x of res as well.

stenzek:
Need to compare skip duplicate frames in advanced options vs present all frames options.

andercard0:
Okay new tests and results are:
 (1x Res)

- Bust a Groove 2 
Skip Duplicate - 24 ~ 28 Tops
Display All Frames 22 ~ 26 Tops

- CTR
Skip Duplicate - 26 ~ 30
Display All frames - 15 ~ 29

- Crash WARPED
Skip Duplicate - 24 ~ 29
Display All frames - 27 ~ 30 
(sound modulating a bit; slow down sound!)

(Warp room 30 fps all time both modes)

stenzek:
What about the "Display All Frames When Possible" option? (that is the "current" default behavior).

Also worth measuring the impact on performance with the software renderer, skipping duplicates will give the sw renderer more time to catch up on the light frames.

andercard0:
> What about the "Display All Frames When Possible" option? (that is the "current" default behavior).

Precisely. That was the case. 
As reported in my previous post but with OpenGl as API used not vulkan or software.

 I can retest with software later.

andercard0:
Tested in Software mode :
(1x Res)

Bust a Groove 2
Display All frames when possible - 9 ~ 29 really playable (some buggy sounds while on menu and during cutscenes)
Skip Duplicate Frames - 12 ~ 29 (some slowdowns but yet playable)

CTR
Display All frames when possible -  21 ~ 29 (on menus and demo mode always 30 if that matters)
Skip Duplicate Frames - 25 ~ 30

Crash WARPED
Display All frames when possible - 22 ~ 29 (some slowdowns here and there)
Skip Duplicate Frames - 27 ~ 29

src/core/host_display.cpp
```diff
@@ -27,6 +28,9 @@ void HostDisplay::SetDisplayMaxFPS(float max_fps)
 
 bool HostDisplay::ShouldSkipDisplayingFrame()
 {
+  if (!m_display_duplicate_frames && !m_display_changed)
+    return true;
+
   if (m_display_frame_interval == 0.0f)
```
ISuckGood, line 34: Shouldnt the members on the side and left side of the equal be swapped. Also i dont think this is the way to compare floating point values  (diff less than machine precision)

stenzek: There's no mathematical operations happening on that value, its loaded straight from config. So an exact comparison is perfectly fine.

stenzek:
This was a bad idea anyway. It doesn't help performance much and using the start address as a heuristic breaks in enough games (e.g ff7 battle screen).

PR #1823, opened 2021-03-16
CDImage: Add support for loading PBP images

ggrtk:
Boots into the first disc by default and doesn't support switching yet. As previously discussed, we'll probably need to extend some CDImage class functionality for that.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -18,6 +18,7 @@ add_library(common
   cd_image_hasher.h
   cd_image_memory.cpp
   cd_image_mds.cpp
+  cd_image_pbp.cpp
   cd_subchannel_replacement.cpp
   cd_subchannel_replacement.h
   cd_xa.cpp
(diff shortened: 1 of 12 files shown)


Blackbird88:
Missing from ALl filetypes?
` "MainWindow",
  "All File Types (*.bin *.img *.iso *.cue *.chd *.ecm *.mds *.exe *.psexe *.psf *.minipsf *.m3u)`

stenzek:
LGTM, many many thanks for all your hard work here!

The big endian check is probably unnecessary, I make no effort to make DuckStation compilable nor runnable on a BE host. But it is in common code.. :)

Still thinking about how to best handle multi-disc, but it'll probably be with some abstraction which can be used for both pbp and playlists, either with an other class in front, using the existing stuff in System, or a "SwitchDisc" method in CDImage. Just not sure if multi-disc stuff belongs in the "CD parsing/reading" library...

PR #1839, opened 2021-03-19
Qt/MemoryCardEditor: Minor update

ggrtk:
Tool tips for making things clearer since people were getting confused I guess.

### DIFF
--- a/src/duckstation-qt/memorycardeditordialog.cpp
+++ b/src/duckstation-qt/memorycardeditordialog.cpp
@@ -62,8 +62,8 @@ void MemoryCardEditorDialog::resizeEvent(QResizeEvent* ev)
 
 void MemoryCardEditorDialog::closeEvent(QCloseEvent* ev)
 {
-  promptForSave(&m_card_a);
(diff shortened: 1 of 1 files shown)


src/duckstation-qt/memorycardeditordialog.cpp
```diff
@@ -30,6 +30,13 @@ MemoryCardEditorDialog::MemoryCardEditorDialog(QWidget* parent) : QDialog(parent
   connectUi();
   populateComboBox(m_ui.cardAPath);
   populateComboBox(m_ui.cardBPath);
+
+  static constexpr char new_card_hover_text[] = TRANSLATABLE("MemoryCardEditorDialog", "New Card...");
```
stenzek, line 34: This can just be `m_ui.newCardA->setTooltip(tr("New Card..."));`

ggrtk: Whoops, not sure why I overcomplicated that... thanks for fixing it.

PR #1840, opened 2021-03-19
CPU/Recompiler: Use MAP_JIT for code space on Apple Silicon

stenzek:
Untested, since I don't own an Intel Mac, let alone an M1 Mac.


MaddTheSane:
Seems to work on my M1 Mac on my OpenEmu Core plug-in.

PR #1883, opened 2021-03-29
Printf-like improvements

MaddTheSane:
* Rename `printflike` to `__printflike`, and use system `__printflike` if available.
* * this is used to make Xcode, the IDE of my choice, show the actual function in drop-down menus instead of just _printflike()_.
* Adds more `__printflike` macros in more places.
* Correct a few improper uses of formatting strings.


MaddTheSane:
It's probably because `__printflike()` is a system macro that Xcode knows what it does.

stenzek:
I don't use Xcode, nor is Mac a supported OS (with no maintainer), so making concessions for it is unreasonable to me.

Other commits look fine.

MaddTheSane:
Let's use #1886 instead.

PR #1954, opened 2021-04-10
(Android) update values-ru/strings

6lackmag3:


alex91093:
Screenshot_20210410-121530_ePSXe
This is what I was talking about in the android support page for specific game touchscreen layouts. This layout works great for megaman legends 2. But for other games it would probably be trash. This layout allows me to turn my character and jump and shoot at the same time.

PR #1972, opened 2021-04-16
Update gamesettings.ini

6lackmag3:


stenzek:
With the new game database, `gamesettings.ini` is no longer used for controller overrides - instead, the supported controllers are listed for each game in the database.

I did a quick check and most of these seem to only have digital controller listed, so should be fine.

6lackmag3:
@stenzek cool. But can we still push fixes for this gamedb? It lists compatible controller types but it doesn't include info on analog/digital mode. E.g. MK4 supports rumble but needs digital mode forcing. How can we workaround this now?

stenzek:
Yup. The gamedb json files are in the repo and split by first letter of the title.

For mk4, you would declare it as only supporting the digital controller. Dualshock forced to digital on start is a special case there, it'll only disable the auto force, not change the type.

PR #1991, opened 2021-04-19

zkdpower:
Update Simple-Chinese language file to latest.
Beside:
The game list column like "Type, Code, Title, File Title" need to adjust width, currently is too narrow. especially in player checking all hooks.
the best expectation is the column adjustable, savable, sequence changable.

some guy tell me change disc would cause soft crash in normal mode unless in NOGui mode.


stenzek:
> The game list column like "Type, Code, Title, File Title" need to adjust width, currently is too narrow. especially in player checking all hooks.

Title and File Title are flexible width, unless you make the window larger it will not have space for both - this is intended.

PR #2028, opened 2021-04-26
Add AppStream metadata file

j8r:
Can be used for AppImage (https://docs.appimage.org/packaging-guide/optional/appstream.html), and required for Flatpak.

Why it is named as `metainfo.xml` instead of `appdata.xml`: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location.

`flatpak run org.freedesktop.appstream-glib validate appimage/org.duckstation.DuckStation.metainfo.xml` is OK, but I still need to test it when building the Flatpak.


stenzek:
Do we want this in the `appimage` directory if it's not appimage-exclusive?

j8r:
I agree, but the things are icons and the desktop file in the appimage directory are also not specific to AppImage.

We can move the appimage script to `scripts/`, and then rename the directory to `dist` (some other projects name like this), or another name.

j8r:
I moved the files, tell me if that's ok for you.

.github/workflows/rolling-release.yml
```diff
@@ -175,7 +175,7 @@ jobs:
         cd build
         cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_NOGUI_FRONTEND=ON -DBUILD_QT_FRONTEND=ON -DUSE_DRMKMS=ON -DUSE_EGL=ON -DUSE_SDL2=ON -DUSE_WAYLAND=ON -DUSE_X11=ON -G Ninja ..
         ninja
-        ../appimage/generate_appimages.sh $(pwd)
+        ../scrtips/generate_appimages.sh $(pwd)
```
stenzek, line 178: Typo?

j8r: right, good catch

dist/org.duckstation.DuckStation.metainfo.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.duckstation.DuckStation</id>
+  <launchable type="desktop-id">org.duckstation.DuckStation.desktop</launchable>
```
j8r, line 4: Side question: do we want duckstation-nogui or duckstation-qt to be the default application (org.duckstation.DuckStation.desktop)? The nogui is quite strange because it has - in fact - a GUI.

Personally I prefer nogui, even though I'm on KDE.

j8r:
I got the Flatpak working, ready to go on my side.

dist/org.duckstation.DuckStation.metainfo.xml
stenzek: Yes. NoGUI is mainly intended for TV-like environments, and doesn't have all the options.

j8r:
One more thing, I noticed a `Terminal=true` in the desktop files. This means, when opening the application launcher, 2 windows open: a terminal + the application.
Is there a reason for this? If not, I suggest to remove the line.

stenzek:
Should be fine to remove.

PR #2030, opened 2021-04-26
Fix width and height checks for fullscreen swapchain (d3d11)

ifarbod:
Height is being compared against Width somehow, this is wrong..


stenzek:
Nice catch, thanks!

PR #2042, opened 2021-05-01

zkdpower:


stenzek:
This is in the incorrect location - it should be in `src/duckstation-qt/translations`.

PR #2046, opened 2021-05-02
Minor changes for macOS

kode54:
The following changes make it possible for me to build an Apple Silicon release, which I can also recursively combine with an x86_64 release, to produce a Universal build, if necessary. It should make life easier for anyone attempting to build it.


stenzek:
Sounds good, many thanks for this! Much appreciated since I can't build it myself.

Creating a universal build sounds good if you know how to do so. I need to make OpenGL optional as well, since Apple's GL driver has bugs everywhere.

I think the MoltenVK build I'm using is a fat dylib, but if needed I can get that rebuilt.

kode54:
For reference, there's a build script here which works for CMake, with some minor changes I patched in in a comment of the PR stream:

https://github.com/dolphin-emu/dolphin/pull/9441

Though I'm not sure if it's possible to install the arm64 version of Homebrew on an x86_64 machine.

stenzek:
Hmm, that is definitely an issue then. I wonder if it's possible to download the arm64 variants of the packages directly. Should only need SDL2 and Qt.

kode54:
Nope. It looks like every CI provider will now need to mass buy M1 Mac mini machines to support this.

https://github.com/Homebrew/brew/issues/10152#issuecomment-831186918

stenzek:
That's why I asked if it's possible to directly download the packages. If we can grab+extract them, then we can just point cmake at the locations (e.g. `-DQt5_DIR`). But that assumes homebrew actually built them for M1 in the first place.

kode54:
Yes, Homebrew's definitions .rb files, which can be viewed or edited with the `brew edit` verb, contain the hashes to the binary packages. The binary locations are pretty much the same for all packages, and just differ in having either big_sur or arm64_big_sur suffix tag.

spotanjo3:
I noticed that you forgot to release the macOS version. I only seen Windows, Linux, Android. No macOS verison.

stenzek:
That's correct. Please see the readme:

> MacOS builds are no longer provided, as I cannot support a platform which I do not own hardware for, and I'm not spending $1000+ out of my own pocket for a machine which I have no other use for.
> 
> You can still build from source, but you will have to debug any issues encountered yourself.
> 
> If anyone is willing to volunteer to support the platform to ensure users have a good experience, I'm more than happy to re-enable the releases.

The CI is still building the mac version, it's just not in the release page, because that implies I'm going to support it. If you really want to, you can download the CI artifact, but I'd strongly recommend building from source yourself, especially if you have an M1 machine.

After I was personally attacked in another thread, I'm definitely not going to pay out of pocket for a cloud mac to debug issues.

spotanjo3:
I understand. And I am sorry about your bad experienced due that you was personally attacked in another thread. That's ridiculous. Anyway, thank you for your time to replying my question. :)

vkedwardli:
> Nope. It looks like every CI provider will now need to mass buy M1 Mac mini machines to support this.

I have successfully cross compile a universal SDL2 on the x86/64 macOS GitHub CI (not optimized yet, just working)
https://github.com/flyinghead/flycast/pull/253/commits/a62b83f301bda3d3ef79c4e2fadc4f4d9b705b3c

PR #2047, opened 2021-05-02

zkdpower:


zkdpower:
"If one of the "separate card per game" memory card **modes** is chosen, these memory cards will be saved to the memcards directory. "this sentence perhaps should change into "If one of the "separate card per game" memory card **type** is chosen, these memory cards will be saved to the memcards directory. "

Issue #1201, opened 2020-12-12
can't run duckstation on windows 7 ultimate 64 bit


Cibomatto2002:
using retroarch 1.9.0 all other emulators I have tried are working for playing ps1 games. mednafen_psx_libretro.dll, pcsx_rearmed_libretro.dll

My laptop may be to old to use duckstation. I don't think it has open gl i know for sure it does not have vulkan. when trying to run duckstation it goes back to retroarch 1.9.0

Acer Aspire 5733Z  I think it cam out in 2011 not sure.

Operating System
	Windows 7 ultimate 64-bit
CPU
	Intel Pentium P6200 @ 2.13GHz
	Arrandale 32nm Technology
RAM
	4.00GB Dual-Channel DDR3 @ 532MHz (7-7-7-20)
Motherboard
	Acer Aspire 5733Z (CPU)
Graphics
	Generic PnP Monitor (1366x768@60Hz)
	Intel HD Graphics (Acer Incorporated [ALI])
Storage
	465GB TOSHIBA MK5059GSXP (SATA )
Optical Drives
	HL-DT-ST DVDRAM GT32N
Audio
	Realtek High Definition Audio

stenzek:
Have you tried the standalone version? You should be able to use the Direct3D renderer, or software (with D3D presentation).

Also, I do not officially support Windows 7 - see the note in the readme:
> Windows 10 is the only version of Windows supported by the developer. Windows 7/8 may work, but is not supported. I am aware some users are still using Windows 7, but it is no longer supported by Microsoft and too much effort to get running on modern hardware. Game bugs are unlikely to be affected by the operating system, however performance issues should be verified on Windows 10 before reporting.

Cibomatto2002:
I had no idea there was a standalone version thanks for telling me I will try it out. i just use retroarch for everything.  iwill try it out and post if it works.

Nucleoprotein:
d3dcompiler_47.dll is in here https://www.microsoft.com/en-us/download/details.aspx?id=35
msvcp140.dll is in here: https://aka.ms/vs/16/release/vc_redist.x64.exe

stenzek:
Closing as this is either a retroarch or support issue. If you still need help, feel free to continue to reply here, or chat to one of us on Discord.

Issue #451, opened 2020-05-16
Can the AppImage be downloaded somewhere?


probonopd:
Thanks for making an AppImage. Can it be downloaded somewhere?

stenzek:
There's a link at the top of the README.md in the repo:
> Latest Windows and Linux (AppImage) Builds: https://ci.appveyor.com/project/stenzek/duckstation

Choose Ubuntu18.04 -> Artifacts.

probonopd:
Thanks, but I must be overlooking something - don't see an AppImage there...

![noappimage](https://user-images.githubusercontent.com/2480569/82119674-f1ac8200-976f-11ea-81f3-4c5a1c5b70a4.png)

ggdrt:
You caught the CI before it had run the AppImage build for that particular commit. It's done now: https://ci.appveyor.com/project/stenzek/duckstation/build/job/qb648n961x0cefhi/artifacts

stenzek:
Sorry, AppVeyor is quite... slow at building.

Perhaps we should write a guide on how to download/use the AppImages like I did for Windows.

ggdrt:
> Perhaps we should write a guide on how to download/use the AppImages like I did for Windows.

Self-assigning this as mentioned in Discord.

probonopd:
Still no AppImage, only a .7z file. Please don't put AppImages into other archives, please see https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages/cc2441518975caca23e9ce2dba6f08a22c678d1e#common-mistake for the reason. Thanks!

stenzek:
There's two appimages, one for each frontend. Both are in the archive. That said I don't see why we couldn't upload them separately though.

ggdrt:
> Please DO NOT put an AppImage into another archive like a .zip or .tar.gz. While it may be tempting to avoid users having to set permission, this breaks desktop integration with the optional appimaged daemon, among other things. 

I don't think breaking desktop integration was really a concern for us since users would just extract the two AppImages out of the archive and overwrite their pre-existing files.

The reason why I archived both together was to have users grab both versions simultaneously like with the Windows builds, but pushing the two AppImages separately seems fine in this case.

stenzek:
Non-zipped AppImages can now be found in the releases page, currently compiled by the GitHub Actions CI: https://github.com/stenzek/duckstation/releases/tag/latest

[...]

PR #2079, opened 2021-05-10
Update installation instructions

j8r:
Adds a condensed `Install` section on top, which is usually what people wants after knowing what is the project.

The new Flatpak installation method is also added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,7 @@
 
 **Discord Server:** https://discord.gg/Buktv3t
 
-**Latest Builds for Windows and Android:** https://github.com/stenzek/duckstation/releases/tag/latest
-
-**Latest Builds for Linux:** [AppImage or Flatpak](#linux)
+**Latest Builds for Windows, Linux (AppImage), and Android** https://github.com/stenzek/duckstation/releases/tag/latest
 
 **Available on Google Play:** https://play.google.com/store/apps/details?id=com.github.stenzek.duckstation&hl=en_AU&gl=US
 
@@ -116,20 +114,6 @@ The Qt frontend includes an automatic update checker. Builds downloaded after 20
 
 ### Linux
 
-#### Flatpak
-
-Works on all Linux distributions with Flatpak installed.
-
-1. [Set up Flatpak](https://www.flatpak.org/setup/)
-
-2. Install DuckStation from [Flathub](https://flathub.org/apps/details/org.duckstation.DuckStation)
-
-`flatpak install -y org.duckstation.DuckStation`
-
-3. Run DuckStation
-
-`flatpak run org.duckstation.DuckStation`
-
 #### Binaries
 
 Prebuilt binaries for 64-bit Linux distros are available for download in the AppImage format. However, these binaries may be incompatible with older Linux distros (e.g. Ubuntu distros earlier than 18.04.4 LTS) due to older distros not providing newer versions of the C/C++ standard libraries required by the AppImage binaries.
@@ -140,6 +124,14 @@ To download:
  - Go to https://github.com/stenzek/duckstation/releases/tag/latest, and download either `duckstation-qt-x64.AppImage` or `duckstation-nogui-x64.AppImage` for your desired frontend.
  - Run `chmod a+x` on the downloaded AppImage -- following this step, the AppImage can be run like a typical executable.
  - Optionally use a program such as [appimaged](https://github.com/AppImage/appimaged) or [AppImageLauncher](https://github.com/TheAssassin/AppImageLauncher) for desktop integration. [AppImageUpdate](https://github.com/AppImage/AppImageUpdate) can be used alongside appimaged to easily update your DuckStation AppImage.
+
+#### [Unofficial] Flatpak
+
+No support for the flatpak package will be provided, any emulator issues must be verified with the AppImage or a source build before being reported.
+
+Flatpak issues can be reported at https://github.com/flathub/org.duckstation.DuckStation/issues.
+
+Follow the instructions at https://github.com/flathub/org.duckstation.DuckStation
 
 ### macOS
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,9 @@
 
 **Discord Server:** https://discord.gg/Buktv3t
 
-**Latest Builds for Windows, Linux (AppImage), and Android** https://github.com/stenzek/duckstation/releases/tag/latest
+**Latest Builds for Windows and Android:** https://github.com/stenzek/duckstation/releases/tag/latest
+
+**Latest Builds for Linux:** [AppImage or Flatpak](#linux)
 
 **Available on Google Play:** https://play.google.com/store/apps/details?id=com.github.stenzek.duckstation&hl=en_AU&gl=US
 
@@ -113,6 +115,22 @@ To set up:
 The Qt frontend includes an automatic update checker. Builds downloaded after 2020/08/07 will automatically check for updates each time the emulator starts, this can be disabled in Settings. Alternatively, you can force an update check by clicking `Help->Check for Updates`.
 
 ### Linux
+
+#### Flatpak
+
+Works on all Linux distributions with Flatpak installed.
+
+1. [Set up Flatpak](https://www.flatpak.org/setup/)
+
+2. Install DuckStation from [Flathub](https://flathub.org/apps/details/org.duckstation.DuckStation)
+
+`flatpak install -y org.duckstation.DuckStation`
+
+3. Run DuckStation
+
+`flatpak run org.duckstation.DuckStation`
+
+#### Binaries
 
 Prebuilt binaries for 64-bit Linux distros are available for download in the AppImage format. However, these binaries may be incompatible with older Linux distros (e.g. Ubuntu distros earlier than 18.04.4 LTS) due to older distros not providing newer versions of the C/C++ standard libraries required by the AppImage binaries.
 


stenzek:
Please don't move the links to the download page and google play listing. People often don't read past the first few lines, and that's critical information.

j8r:
Maybe but there is too much information to fit, or there are ones lacking.
See https://github.com/matiassingers/awesome-readme for examples.

Some projects have a link summary on the top, which I find more condensed/better.

stenzek:
I'm not budging on this. The users of this application are much less technically literate than users of libraries, and the less stupid-friendly it is, the more support requests I have to deal with.

j8r:
Exactly, that's why the lack of explanations about the links can be a problem: how to use this archives?

j8r:
Only have a clear header referencing the install instructions is not good? It will be just one click away, and will provide more help for the newcomers vs throwing a link.

j8r:
Yes, that's why I removed the links. The links I removed are superfluous, and lacks additional information for newcomers to make the emulator work.

stenzek:
Sorry, you're not the one who has to deal with the user's support requests. I'm not changing it.

Feel free to add the flatpak instructions to the Linux installation section, but I'm not merging this in its current state.

j8r:
Ok. Do we want a cleaner, simpler installation section? I find the current one has to much explanations, details - usually we just want to get it installed, then configure it.

Maybe having subsections inside the current Download and running?

j8r:
Also, not sure why some links are in the header (like Download and running, features) and some below as links (discord server, wiki).

CookiePLMonster:
> Also, not sure why some links are in the header (like Download and running, features) and some below as links (discord server, wiki).

> People often don't read past the first few lines, and that's critical information.

Discord/wiki links are less important for the end users than the download links.

j8r:
> Discord/wiki links are less important for the end users than the download links.

They are in bold an occupy each one line, so they are visually more important than the header links :confused: 

Anyway, this may be just me being confused. I'd just add a section.

stenzek:
If you want to take over all user support across all platforms, then you can rewrite the readme how you see fit.

But as it stands I see no compelling reason to change any of the structure.

j8r:
As you wish. 

I thought it could be improved because of:
- https://github.com/stenzek/duckstation/issues/451#issuecomment-629633851
- https://github.com/stenzek/duckstation/issues/1201#issuecomment-743691225

But one can say this are people not reading the README rather being that not clear.

README.md
```diff
@@ -3,7 +3,9 @@
 
 **Discord Server:** https://discord.gg/Buktv3t
 
-**Latest Builds for Windows, Linux (AppImage), and Android** https://github.com/stenzek/duckstation/releases/tag/latest
+**Latest Builds for Windows and Android** https://github.com/stenzek/duckstation/releases/tag/latest
+
+**Linux** https://github.com/stenzek/duckstation#linux
```
j8r, line 8: I've created a link here to reference both the AppImage and Flatpak installation methods. The later is put forward because it is cross-distribution.

stenzek:
Upon further thought, I've decided I do not want the flatpak linked in the main page. I have no control over it, and putting it here gives the users it's official, and therefore I have to provide support for it.

j8r:
That's not true: you have access to https://github.com/flathub/org.duckstation.DuckStation.

Flatpak should have less issues, at least in the long-run, than other alternatives like building locally or the AppImage, because the environment is fully controlled and packaged.

Plus, nothing stops you to ping me if an issue arise - no solution is perfect.

stenzek:
I don't have control over the build process. That's what I need, especially with the (very soon) move to Qt6.

stenzek:
Why are we bothering with this mess, when we could just build it in the CI and offer it for download here then?

j8r:
The result would be the same. We either use GitHub Actions or the Flathub CI. One difference is the publishing to the Flathub store.

stenzek:
I don't care about flathub stores.

stenzek:
Does a user need to install some flatpak package manager to use this package? Or can they simply download and run it? That's the simple solution I want.

Remember I said way back when this was first proposed that I'm okay with third-party packages, but only if they are not marked as official (except not really, but I can't do much about them I mistakenly made this software open-source). Because if it's official, users expect me to provide support for it.

j8r:
Yes you have to have `flatpak` installed.

What we can do is add a link to https://github.com/flathub/org.duckstation.DuckStation, have the instructions there, and then mention that any issue related to the Flatpak bundle should be reported in this very repository.

Of course an issue unrelated to Flatpak will indeed end-up here.

j8r:
I agree about statically linked binaries. AppImage won't add any benefit (AFAIK). However, it doesn't exclude to have an unofficial Flatpak. I've experience on this, in case you need help.

`flatpak` is installed by default on Fedora, on officially supported by Red Hat on RHEL.
`snap` is installed by default on Ubuntu, and of course officially supported.

j8r:
Why do you want to add the Flatpak build in the GitHub release? I don't see any advantage.

I'll change the Flatpak repository in order to get the commit hash from the `latest` tag instead of master.

stenzek:
Perhaps I misunderstood your intent to replace AppImages with Flatpaks then.

1. Like I said, I want a build environment I can control, that way if there's issues I can reproduce them. Using third party builders does not give me that.
2. A third-party download source which someone other than myself is a potential security vulnerability.

Thus, it makes sense to have one source for downloads; GitHub releases.

stenzek:
Anyway, I'm sorry for wasting your time here. It's clear that I misunderstood how the flatpak process would work - I was under the impression it was the same as AppImages and we could eventually migrate over with very minimal procedural changes.

As best I can tell, the appimage has unfixable flaws, so going static is probably the right way for official downloads to meet my requirements. Hopefully it doesn't result in too massive files.... Will probably look into that around the same time I do the Qt 6 migration.

j8r:
What's doing remotely and locally should be identical, like here if CMake is ran in GitHub actions or if you run it locally.

A bit like how Docker works, building locally of by any CI must result to the same hash.

I've to check this to have the proof for Flatpak. What's sure is builds are reproducible, but I'm not 100% sure if cloud builds are the same as local ones.

Have you checked if your local builds are the same to the GitHub Actions one? They may differ due to different CPU instructions support.

j8r:
> A third-party download source which someone other than myself has access to, is a potential security vulnerability.

What do you mean? You can remove me to access the repository.

Flathub members can indeed modify the repository, like (some) GitHub employees can modify any repository.

stenzek:
> What's doing remotely and locally should be identical, like here if CMake is ran in GitHub actions or if you run it locally.

Do they use an Ubuntu 20.04 environment? Because that's what I'm working with.

> What do you mean? You can remove me to access the repository.

I don't have time to maintain another package and work out all the intricies myself.

> Flathub members can indeed modify the repository, like (some) GitHub employees can modify any repository.

I trust GitHub more than flathub, not saying they're malicious, but it's difficult to trust something I don't use myself.

There's also another issue - new features added to duckstation are now encrypted. The base emulator is staying fully open source, but due to people commercializing and selling it, I've been forced to keep any new features out of their hands (e.g. the reintroduced libretro core). This requires a key to decrypt as part of the build process, and I want to keep access to these keys locked down (i.e. GitHub secrets only). I'm really not happy about this at all, but I don't see any other option :(

j8r:
The distribution is irrelevant, here the "packages" - at least libraries, at from the SDK. So, Ubuntu or not, any distribution will be the same ones. Think like Docker, a Flatpak is very like a Docker image.

I've updated this PR, but we can't see the change because of being closed. What it brings:

> #### Flatpak
> **Any issue related to the package it-self should be reported [here](https://github.com/flathub/org.duckstation.DuckStation/issues)**
>
> Follow the instructions at https://github.com/flathub/org.duckstation.DuckStation

j8r:
>  I've been forced to keep any new features out of their hands (e.g. the reintroduced libretro core). This requires a key to decrypt as part of the build process, and I want to keep access to these keys locked down (i.e. GitHub secrets only). I'm really not happy about this at all, but I don't see any other option :(

I not particularly fond of this, because we can't even less trust you - the build is not reproducible at all.
You could have two editions: a FOSS one and a commercial one.

stenzek:
Absolutely. I really dislike the idea too, and it makes things more complex for building too. But my hands are tied, I'm not comfortable with these people making any more money off the work I give away for free. Also it's not commerical, the opposite in fact, it states "commerical distribution is prohibited" :)

> The distribution is irrelevant, here the "packages" - at least libraries, at from the SDK. So, Ubuntu or not, any distribution will be the same ones. Think like Docker, a Flatpak is very like a Docker image.

I see, thanks for the explanation. In that case, maybe it would simplify qt 6 packaging if I can profile my own base image?

README.md
```diff
@@ -3,7 +3,9 @@
 
 **Discord Server:** https://discord.gg/Buktv3t
 
-**Latest Builds for Windows, Linux (AppImage), and Android** https://github.com/stenzek/duckstation/releases/tag/latest
```
stenzek, line 6: Leave these lines as is

README.md
```diff
@@ -114,6 +116,14 @@ The Qt frontend includes an automatic update checker. Builds downloaded after 20
 
 ### Linux
 
+#### Flatpak
+
```
stenzek, line 120: Change this to "No support for the flatpak package will be provided, any emulator issues must be verified with the appimage or a source build before being reported. Flatpak issues can be reported at: ...."

And move it underneath the first link.

j8r:
I'm ok to make it unofficial at first. With time, when we'll be more confident, and the bundle being more stable, the status could eventually change.

stenzek:
Building qt with every release would take way too long.

I'm not depending on any other sources. The out of state qt version is the reason for half of the issues with the appimage.

> I'm ok to make it unofficial at first. With time, when we'll be more confident, and the bundle being more stable, the status could eventually change.

It's basically a third party package which you're asking me to advertise at the moment, since I don't use this package manager.

j8r:
Qt6 will solve all the Qt5 issues, how it is related to AppImage specifically?

stenzek:
The appimage was using Ubuntu 18.04 and qt5.9, which had broken hidpi scaling among other issues. I think 20.04 uses 5.12, but it's too old is missing a few things (have version checks in the code)

j8r:
Here it is Qt 5.15, so the latest LTS stable. I doubt moving to Qt 6 will solve all issues. It will solve some, and create others.

stenzek:
No. Nogui is a cut down, full screen UI intended for TVs. It's missing quite a few features in the qt frontend (e.g. memory card editor, cheat manager, debugger, game properties, etc)

j8r:
Yeah, you said that previously. I mean, bring this features to the nogui edition? It is of course lots of work, I'm only thinking if in the long run it could be better, vs. having two editions, with one potentially more broken.

PR #2080, opened 2021-05-10
Widescreen Hotkey

ValadAmoleo:
Added the ability to toggle Widescreen Hack and set the Aspect Ratio via a hotkey.

This is in response to Discussion #1912.

### DIFF
--- a/src/frontend-common/common_host_interface.cpp
+++ b/src/frontend-common/common_host_interface.cpp
@@ -12,6 +12,7 @@
 #include "core/cpu_code_cache.h"
 #include "core/dma.h"
 #include "core/gpu.h"
+#include "core/gte.h"
 #include "core/host_display.h"
 #include "core/mdec.h"
 #include "core/pgxp.h"
@@ -2039,6 +2040,14 @@ void CommonHostInterface::RegisterGraphicsHotkeys()
                      g_texture_replacements.Reload();
                    }
                  });
+
+  RegisterHotkey(StaticString(TRANSLATABLE("Hotkeys", "Graphics")), StaticString("ToggleWidescreen"),
+                 StaticString(TRANSLATABLE("Hotkeys", "Toggle Widescreen")), [this](bool pressed) {
+                   if (pressed && System::IsValid())
+                   {
+                     ToggleWidescreen();
+                   }
+                 });
 }
 
 void CommonHostInterface::RegisterSaveStateHotkeys()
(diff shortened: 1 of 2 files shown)


src/frontend-common/common_host_interface.cpp
```diff
@@ -3344,6 +3353,18 @@ void CommonHostInterface::ReloadPostProcessingShaders()
     AddOSDMessage(TranslateStdString("OSDMessage", "Post-processing shaders reloaded."), 10.0f);
 }
 
+void CommonHostInterface::ToggleWidescreen()
+{
+  g_settings.gpu_widescreen_hack = !g_settings.gpu_widescreen_hack;
```
stenzek, line 3358: Would it be better to switch between the chosen AR and native here, rather than 16:9?

ValadAmoleo: You're right, of course.  I've pushed another commit so that it does this.

src/frontend-common/common_host_interface.cpp
```diff
@@ -3344,6 +3353,43 @@ void CommonHostInterface::ReloadPostProcessingShaders()
...
+  const GameSettings::Entry* gs = m_game_list->GetGameSettings(System::GetRunningPath(), System::GetRunningCode());
+  DisplayAspectRatio userRatio;
+  if (gs && gs->display_aspect_ratio.has_value())
+    userRatio = gs->display_aspect_ratio.value();
+  else
+  {
+    Settings o_settings;
```
stenzek, line 3365: Doesn't need to go through the whole settings load process, see https://github.com/stenzek/duckstation/blob/41cf93793dd0e6a0b11b1f1b68243b0149f4c943/src/core/settings.cpp#L225-L228

src/frontend-common/common_host_interface.cpp
```diff
@@ -3344,6 +3354,31 @@ void CommonHostInterface::ReloadPostProcessingShaders()
...
+          ->GetStringValue("Display", "AspectRatio", Settings::GetDisplayAspectRatioName(DisplayAspectRatio::Auto))
+        .c_str())
+        .value_or(DisplayAspectRatio::Auto);
+  if (userRatio == DisplayAspectRatio::Auto || userRatio == DisplayAspectRatio::PAR1_1 || userRatio == DisplayAspectRatio::R4_3)
+    g_settings.display_aspect_ratio = g_settings.gpu_widescreen_hack ? DisplayAspectRatio::R16_9 : userRatio;
+  else
+    g_settings.display_aspect_ratio = g_settings.gpu_widescreen_hack ? userRatio : DisplayAspectRatio::Auto; 
+  
+  String arMessage;
+  arMessage.AppendFormattedString("Widescreen Hack is now %s and aspect ratio set to %s.", g_settings.gpu_widescreen_hack ? "enabled" : "disabled", Settings::GetDisplayAspectRatioName(g_settings.display_aspect_ratio));
```
stenzek, line 3376: This is not translatable, but I'll fix it up later.

PR #2106, opened 2021-05-12
Added a toggle for persistent message.

ValadAmoleo:
Added a toggle to display settings to disable persistent messages (ff, pause) in the fullscreen UI.

I don't think anybody has been shouting for this but while using the emulator yesterday I realised that I play with turbo enabled quite often and in the fullscreen UI (only) this means there's a permanent on screen icon.  Which is great so that the user knows what's going on but bad if you're playing on a display prone to burn in (e.g. OLED).  So I added this to the options under display.

Defaults to true, which is existing functionality.

Totally understand if you decline this as it's not going to be a widespread problem.  Also I'm not sure on if "persistent messages" is the most user friendly terminology.  And, of course, this is only applicable to the fullscreen UI as there are no persistent messages in the qt desktop version.

### DIFF
--- a/src/core/host_interface.cpp
+++ b/src/core/host_interface.cpp
@@ -542,6 +542,7 @@ void HostInterface::SetDefaultSettings(SettingsInterface& si)
   si.SetBoolValue("Display", "Stretch", false);
   si.SetBoolValue("Display", "PostProcessing", false);
   si.SetBoolValue("Display", "ShowOSDMessages", true);
+  si.SetBoolValue("Display", "ShowPersistentMessages", true);
   si.SetBoolValue("Display", "ShowFPS", false);
   si.SetBoolValue("Display", "ShowVPS", false);
   si.SetBoolValue("Display", "ShowSpeed", false);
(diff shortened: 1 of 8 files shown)


stenzek:
I really disagree with this idea. But if you must have it, it's probably best placed in advanced options, display settings are already to cluttered and the 768p gang will complain if it's any larger.

ValadAmoleo:
Moved the option to the advanced options page.

If you do disagree with this though, just decline the pull request.  I thought I'd at least offer it up as it just makes the fullscreen UI behave like the qt one as far as these icons are concerned.  Only did it for my personal use because I play on an OLED TV and wanted to try out the achievements feature (which requires the fullscreen UI).

stenzek:
I meant in the advanced option list, not outside of it. But that's okay, I'll move it. I think "Show Status Indicators" is a better name.

The main reason I dislike it is because when it's paused and you're fullscreen, there's zero indication that it's actually paused. So you might think it's frozen. When it's windowed you have the toolbar/status bar indicating it's paused.

PR #2135, opened 2021-05-19

andercard0:
Update file to latest.


stenzek:
Manually merged in https://github.com/stenzek/duckstation/commit/2a3b3960dcfd70eff96f3ce59176a181c76d5a12 because history got clobbered.

PR #2138, opened 2021-05-19
(Android) - Atualização Português do Brasil

andercard0:
Update String file to latest.


stenzek:
Manually merged in https://github.com/stenzek/duckstation/commit/c417bd7a355832d6c682de34f280e3566d7d2c81 because history got clobbered.

PR #2202, opened 2021-06-01
Update Russian translation

shikulja:
Sorry for the old commits. It looks like github has changed something again.


stenzek:
All good, I can squash them.

PR #2213, opened 2021-06-04
Script improvements

Margen67:


stenzek:
These scripts aren't meant to be in the main repository; I've removed them.

PR #2215, opened 2021-06-04
Add an "Increase Safe Zone" option to move OSD messages further away from screen edges

CookiePLMonster:
Useful for phones with rounded edges, as currently OSD messages are being cut off.

Untested, let's see what comes out of CI.


Calinou:
Should a safe zone be always enabled (or at least the option be enabled by default)? I think this is a more universal option.

stenzek:
Implemented in https://github.com/stenzek/duckstation/commit/056b62983787ab338f8aa9220713f921aef634c1.

PR #2224, opened 2021-06-05
GPU: Implement texture dumping and replacement

stenzek:
This is still a work in progress, but I'm PR'ing it so anyone interested can mess with it.

Rough TODO:

- [x] Support dumping by VRAM write and clut/palette.
- [x] Support replacing textures.
- [ ] Improve dumping further and reduce the number of junk textures (it's much better than the initial version, but there's still room for improvement.
- [x] Implement on backends other than DX11.
- [x] Make sure it doesn't crash the software renderer.
- [ ] Save texture replacement state to save state (at the moment it will be lost when you load state).
- [ ] Save texture dumper state to save state (same as replacements, it's lost when loading).
- [ ] Texture filtering options for the replacements (mipmaps/trilinear).

NOTE: This is a completely different/original implementation when compared to the functionality in beetle-psx. Texture packs created for it are not, and never will be compatible.

Settings:
img

Some initial results:
https://cdn.discordapp.com/attachments/792256355280355336/848559538672369664/Tomb_Raider_USA_2021-05-30_23-51-10.mp4
https://cdn.discordapp.com/attachments/695640472202379428/848895332749148190/Castlevania_-_Symphony_of_the_Night_USA_2021-05-30_19-47-52.mp4

img
img
img
img
img
img
img
img

### DIFF
--- a/src/common/image.cpp
+++ b/src/common/image.cpp
@@ -186,7 +186,7 @@ void RGBA8Image::Resize(u32 new_width, u32 new_height)
   SetPixels(new_width, new_height, std::move(resized_texture_data));
 }
 
-void RGBA8Image::Resize(const RGBA8Image* src_image, u32 new_width, u32 new_height)
+void RGBA8Image::ResizeFrom(const RGBA8Image* src_image, u32 new_width, u32 new_height)
 {
   if (src_image->m_width == new_width && src_image->m_height == new_height)
   {
--- a/src/common/image.h
+++ b/src/common/image.h
@@ -57,6 +57,11 @@ public:
   ALWAYS_INLINE void SetPixel(u32 x, u32 y, PixelType pixel) { m_pixels[y * m_width + x] = pixel; }
   ALWAYS_INLINE PixelType GetPixel(u32 x, u32 y) const { return m_pixels[y * m_width + x]; }
 
+  ALWAYS_INLINE typename std::vector<PixelType>::const_iterator begin() const { return m_pixels.begin(); }
+  ALWAYS_INLINE typename std::vector<PixelType>::const_iterator end() const { return m_pixels.end(); }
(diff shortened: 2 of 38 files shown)


andercard0:
Suggestions:
- Shortcuts to enable and disable texture replacement
i.e: F11 - To turn On and Off on the go will be great to save time while testing and see the results. 

- Tooltips: 
There are too many options in Texture Replacement Settings menu I'm not sure what all of those do so my suggestion is to create some tooltips on mouse hover, to briefly shortly which one does what, not all, only the category group is for will be okay-ish.

stenzek:
Turning it off and on at runtime isn't possible - it needs the VRAM upload as it happens. You could temporarily disable it, but fully disabling would not be recoverable.

The dumping options aren't meant for users who don't know what they're doing. That's why it's in a separate dialog away from the other settings. I'll document it on the wiki some time.

andercard0:
> 
> The dumping options aren't meant for users who don't know what they're doing. That's why it's in a separate dialog away from the other settings. I'll document it on the wiki some time.

Cool then, what would be great!

andercard0:
open dump directory isn't opening, just dump and change some here where should i put those ?

danprice142:
This is awesome! Cant wait for this feature on Duckstation and PCSX2 to hit the core versions. Great work and Thanks so much.

intra32:
Will there be support for mipmaps? (and if so, any chance to upload textures in a format supporting custom mipmaps, e.g. DDS?)

tvhung83:
Super excited when I heard about this, thank you so much for your hard work!

stenzek:
No point keeping this open, the approach was flawed. I'm considering other options, but it's a challenging problem, since there is no accurate source of truth for texture sizes.

PR #2238, opened 2021-06-08
Misc improvements

CookiePLMonster:
A bunch of loosely related improvements:

* Hide the `Load` button in the save state UI when playing in Hardcore mode (as loading save states is impossible).
* Disallow underclocking in Hardcore mode. Any means of slowing the game down are supposed to be blocked, but the user could still try to slow the game down by underclocking the emulated console. Now clock speeds below 100% are forbidden, just like it was already done with game speed settings (faster is OK, slower is not).
* Disallow loading the save state and/or the resume save state via the command line in Hardcore mode. This could be considered an exploit, so it needs to go.
* Unify CPU clock speed behaviour in Game Properties and Console Settings - now the CPU clock slider is greyed out when overclocking is disabled, and the clock speed shows as 100% when overclocking is disabled.

### DIFF
--- a/src/core/host_interface.cpp
+++ b/src/core/host_interface.cpp
@@ -83,14 +83,14 @@ s32 HostInterface::GetAudioOutputVolume() const
   return g_settings.audio_output_muted ? 0 : g_settings.audio_output_volume;
 }
 
-bool HostInterface::BootSystem(const SystemBootParameters& parameters)
+bool HostInterface::BootSystem(std::shared_ptr<SystemBootParameters> parameters)
 {
-  if (!parameters.state_stream)
+  if (!parameters->state_stream)
   {
-    if (parameters.filename.empty())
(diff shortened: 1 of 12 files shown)


src/duckstation-qt/main.cpp
```diff
@@ -103,6 +104,13 @@ int main(int argc, char* argv[])
 
   if (boot_params)
   {
+    // In Challenge mode, do not allow loading a save state under no circumstances
+    // If it's present, drop it
+    if (host_interface->IsCheevosChallengeModeActive())
+    {
+      boot_params->state_stream.reset();
+    }
```
CookiePLMonster, line 112: Perhaps this should be done in `bootSystem` to make it absolutely bulletproof? Other code paths shouldn't be passing the state stream to `bootSystem` but it could act as an additional safeguard.

PR #2240, opened 2021-06-08
Two bugfixes and slight refactor of the boot parameter code as a followup to #2210

mgziminsky:
Don't open resume file during command line parsing, do it from the scope that calls LoadState in order to minimize the scope of filesystem locks.
Stopped using pointers for boot params since they aren't shared and are now all stack data.

Up to you if you actually want to merge this since it's just meant as minor code cleanup. The two smaller commits are fixing actual bugs though, so if you don't want the larger refactor, you can cherry-pick those two, or I can submit a separate PR if you prefer.

### DIFF
--- a/src/core/host_interface.cpp
+++ b/src/core/host_interface.cpp
@@ -914,8 +914,7 @@ void HostInterface::CheckForSettingsChanges(const Settings& old_settings)
 
 void HostInterface::SetUserDirectoryToProgramDirectory()
 {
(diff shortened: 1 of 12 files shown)


stenzek:
The state being a stream and not a filename was a deliberate design decision, so that the frontend could add extra metadata (e.g. retroarchivements) to the state, or load from memory. So I'm not really a fan of changing that.

stenzek:
Closing as the refactoring has made most of these changes redundant from a quick glance.

PR #2243, opened 2021-06-09
Add an "Unofficial Test Mode" to allow unlocking non-core achievements

CookiePLMonster:
As in the title, this PR adds an option to unlock local/unofficial achievements without sending them to the backend server. This behaviour corresponds to an "official" topic about unofficial achievements on RetroAchievements boards:
https://retroachievements.org/viewtopic.php?t=6715

Option not available on Android at the moment, but maybe it should be added?
image
image

### DIFF
--- a/src/frontend-common/cheevos.cpp
+++ b/src/frontend-common/cheevos.cpp
@@ -55,6 +55,7 @@ u32 g_game_id = 0;
 
 static bool s_logged_in = false;
 static bool s_test_mode = false;
+static bool s_unofficial_test_mode = false;
 static bool s_use_first_disc_from_playlist = true;
 static bool s_rich_presence_enabled = false;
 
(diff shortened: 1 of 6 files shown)


src/frontend-common/cheevos.cpp
```diff
@@ -1101,6 +1110,13 @@ void UnlockAchievement(u32 achievement_id, bool add_notification /* = true*/)
     return;
   }
 
+  if (achievement->category != AchievementCategory::Core)
+  {
+    Log_WarningPrintf("Skipping sending achievement %u unlock to server because it's not from the core set.",
```
stenzek, line 1115: Maybe it'd be worth including this somewhere in the unlock message? (so the user doesn't go looking online for it)

CookiePLMonster: So, append the information that it's an unofficial achievement somewhere in the unlock pop-up?

stenzek: Sure, that works

PR #2251, opened 2021-06-10
Fix a crash when shutting down emulation from the Hardcore mode confirmation prompt

CookiePLMonster:
This fixes a crash after the following steps:

1. Run DuckStation and start the game. Fullscreen UI has to be enabled, either by running the game from a NoGUI executable or by having the Fullscreen UI enabled for the Qt frontend.
2. Go to Achievement Settings and enable Hardcore Mode.
3. Say Yes when prompted for confirmation.
4. DuckStation crashes.

### DIFF
--- a/src/frontend-common/fullscreen_ui.cpp
+++ b/src/frontend-common/fullscreen_ui.cpp
@@ -1203,6 +1203,7 @@ static bool ConfirmChallengeModeEnable()
   if (!s_host_interface->ConfirmMessage(message))
     return false;
 
+  SaveAndApplySettings();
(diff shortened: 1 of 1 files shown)


stenzek:
I think simply moving the `SaveAndApplySettings()` into `ConfirmChallengeModeEnable()` would be a cleaner solution.

CookiePLMonster:
> I think simply moving the `SaveAndApplySettings()` into `ConfirmChallengeModeEnable()` would be a cleaner solution.

Sounds good, will amend later today.

PR #2302, opened 2021-06-23
EvdevControllerInterface: Actually use controller_id

tmaillart:
Multiple controller was not possible using evdev backend.


tmaillart:
I did not see there was a fix, sorry

PR #2345, opened 2021-07-02

zkdpower:

besides:
new version added function - "Excluded Paths" only can add files, can't add directory.

in my point of view, "Search Directories"- "Path" should change into "Search Directories"- "Directories".
then could show the "directory" feature to add directory, show the "path" to add single file. or else,  the previous path can add directory, the next path can not add directory would arouse a contradiction.


stenzek:
> new version added function - "Excluded Paths" only can add files, can't add directory.

This is intentional. It's a way to remove duplicate scanned games from the list (e.g. if you have a playlist, and want to remove the individual games from the list, hence exclude)

PR #2351, opened 2021-07-03

zkdpower:

Besides:
if i Resume any saved game, and load state from "Game Save" slot, not from "Global Save" slot. the "Undo load state" menu won't active until i try to save "Game Save" state once,  if without "Game Save" saving operation, the "Undo load state" menu will never lightened, i think it is a defect.  it is supposed to be lightenable directily after loading any state twice. because loading any "Global Save" twice can use the undo function.

in fact, although the menu "Undo load state" button is often unlightened，but undo function is active normally, because the hot key which bound the "Undo load state" still can works ,just out-sync with menu.

renaming "Undo load state" to "Undo the last load state" is more clear for user, because once the "Undo load state" button is on, it will not go out, only "Undo the last load state" can reasonable when click it doesn't contine to undo, "Undo load state" keeps lightened but can't continue to undo is unreasonable.

the ideal design is once the undo command successfully, the menu sync extinguished, and the hot key function sync ineffective.
they won't works until next successful loading.

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_zh-cn.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_zh-cn.ts
@@ -312,25 +312,25 @@ Login token generated on %2.</source>
     </message>
     <message>
         <location filename="../advancedsettingswidget.ui" line="64"/>
-        <location filename="../advancedsettingswidget.cpp" line="203"/>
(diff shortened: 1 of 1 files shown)


stenzek:
Desync between the menu and internal state fixed in https://github.com/stenzek/duckstation/commit/3e7501c5c8773c3617a33fd3cd72dea5bc903e5d

PR #2438, opened 2021-07-23
GPU/Vulkan: Add debug utility functions

Wunkolo:
Adds a way to set the name of vulkan objects, and automatically determining
their `VkObjectType` enum value at compile-time. As well as a utility-object
for creating debug-scopes for `VkQueue` and `VkCommandBuffer` objects.
`DebugScope` objects will automatically emit `Begin` and `End` commands
within the ctor and dtor and allow for C++ scopes to name and color
sections of a vulkan command. These DebugScopes are also able
to be nested within each other and will automatically pick a color
depending on the current recursive depth.

These functions are all null-stubbed in non-debug compilations.

before:
This is how a captured command buffer would look like in renderdoc and nsight
before
before-nsight

after:
Now we have the exact C++-side scope of the function emitting the particular commands for each particular queue being submitted to, as well as profiling data for each scope
after
after-nsight


src/common/vulkan/util.h
```diff
@@ -74,10 +74,141 @@ VkShaderModule CompileAndCreateFragmentShader(std::string_view source_code);
...
+
+template<typename T>
+inline void SetObjectName(VkDevice device, T object_handle, const char* object_name)
+{
+  SetObjectName(device, reinterpret_cast<void*>((typename VkObjectTypeMap<T>::type)object_handle),
+                VkObjectTypeMap<T>::value, object_name);
+}
+
+// Command buffer debug utils
+void BeginDebugScope(VkCommandBuffer commandBuffer, const char* scope_name,
```
stenzek, line 130: It might be better to just define these in the .h rather than the .cpp, with duplicate definitions (error prone if one side of the ifdef is updated but not the other).

src/common/vulkan/util.h
```diff
@@ -74,10 +74,141 @@ VkShaderModule CompileAndCreateFragmentShader(std::string_view source_code);
...
+template<> struct VkObjectTypeMap<VkImage					> { using type = VkImage					; static constexpr VkObjectType value = VK_OBJECT_TYPE_IMAGE; };
+template<> struct VkObjectTypeMap<VkEvent					> { using type = VkEvent					; static constexpr VkObjectType value = VK_OBJECT_TYPE_EVENT; };
+template<> struct VkObjectTypeMap<VkQueryPool				> { using type = VkQueryPool				; static constexpr VkObjectType value = VK_OBJECT_TYPE_QUERY_POOL; };
+template<> struct VkObjectTypeMap<VkBufferView				> { using type = VkBufferView				; static constexpr VkObjectType value = VK_OBJECT_TYPE_BUFFER_VIEW; };
+template<> struct VkObjectTypeMap<VkImageView				> { using type = VkImageView				; static constexpr VkObjectType value = VK_OBJECT_TYPE_IMAGE_VIEW; };
+template<> struct VkObjectTypeMap<VkShaderModule			> { using type = VkShaderModule				; static constexpr VkObjectType value = VK_OBJECT_TYPE_SHADER_MODULE; };
+template<> struct VkObjectTypeMap<VkPipelineCache			> { using type = VkPipelineCache			; static constexpr VkObjectType value = VK_OBJECT_TYPE_PIPELINE_CACHE; };
```
stenzek, line 104: Tabs instead of spaces

src/common/vulkan/util.h
```diff
@@ -74,10 +74,141 @@ VkShaderModule CompileAndCreateFragmentShader(std::string_view source_code);
...
+template<> struct VkObjectTypeMap<VkFramebuffer				> { using type = VkFramebuffer				; static constexpr VkObjectType value = VK_OBJECT_TYPE_FRAMEBUFFER; };
+template<> struct VkObjectTypeMap<VkCommandPool				> { using type = VkCommandPool				; static constexpr VkObjectType value = VK_OBJECT_TYPE_COMMAND_POOL; };
+template<> struct VkObjectTypeMap<VkDescriptorUpdateTemplate> { using type = VkDescriptorUpdateTemplate	; static constexpr VkObjectType value = VK_OBJECT_TYPE_DESCRIPTOR_UPDATE_TEMPLATE; };
+template<> struct VkObjectTypeMap<VkSurfaceKHR				> { using type = VkSurfaceKHR				; static constexpr VkObjectType value = VK_OBJECT_TYPE_SURFACE_KHR; };
+template<> struct VkObjectTypeMap<VkSwapchainKHR			> { using type = VkSwapchainKHR				; static constexpr VkObjectType value = VK_OBJECT_TYPE_SWAPCHAIN_KHR; };
+template<> struct VkObjectTypeMap<VkDebugUtilsMessengerEXT	> { using type = VkDebugUtilsMessengerEXT	; static constexpr VkObjectType value = VK_OBJECT_TYPE_DEBUG_UTILS_MESSENGER_EXT; };
+// clang-format on
+
+void SetObjectName(VkDevice device, void* object_handle, VkObjectType object_type, const char* object_name);
```
stenzek, line 120: Any reason why these can't be varargs? (avoids the heap allocation for strings, which can be slow in debug builds)

stenzek (commented):
Thanks for this! Personally I manage without any labels (usually go off the shaders/pipeline), but if it helps for others, that's great.

src/common/vulkan/util.h
```diff
@@ -74,10 +76,226 @@ VkShaderModule CompileAndCreateFragmentShader(std::string_view source_code);
...
+template<> struct VkObjectTypeMap<VkCommandPool             > { using type = VkCommandPool             ; static constexpr VkObjectType value = VK_OBJECT_TYPE_COMMAND_POOL;               };
+template<> struct VkObjectTypeMap<VkDescriptorUpdateTemplate> { using type = VkDescriptorUpdateTemplate; static constexpr VkObjectType value = VK_OBJECT_TYPE_DESCRIPTOR_UPDATE_TEMPLATE; };
+template<> struct VkObjectTypeMap<VkSurfaceKHR              > { using type = VkSurfaceKHR              ; static constexpr VkObjectType value = VK_OBJECT_TYPE_SURFACE_KHR;                };
+template<> struct VkObjectTypeMap<VkSwapchainKHR            > { using type = VkSwapchainKHR            ; static constexpr VkObjectType value = VK_OBJECT_TYPE_SWAPCHAIN_KHR;              };
+template<> struct VkObjectTypeMap<VkDebugUtilsMessengerEXT  > { using type = VkDebugUtilsMessengerEXT  ; static constexpr VkObjectType value = VK_OBJECT_TYPE_DEBUG_UTILS_MESSENGER_EXT;  };
+// clang-format on
+
+inline void SetObjectName(VkDevice device, void* object_handle, VkObjectType object_type, const char* format, ...)
+#ifdef _DEBUG
```
stenzek, line 121: Stylistic nitpick; put the `ifdef` after the braces.

src/common/vulkan/util.h
```diff
@@ -74,10 +76,226 @@ VkShaderModule CompileAndCreateFragmentShader(std::string_view source_code);
...
+  vkCmdInsertDebugUtilsLabelEXT(command_buffer, &label);
+}
+#else
+{
+}
+#endif
+
+// Queue debug utils
+inline void BeginDebugScope(VkQueue queue, const char* scope_name,
+                            std::array<float, 4> scope_color = {0.75, 0.75, 0.75, 1.0})
```
stenzek, line 213: Pass `scope_color` by const reference, it won't fit in a register.

src/core/gpu_hw_vulkan.cpp
```diff
@@ -780,12 +848,23 @@ void GPU_HW_Vulkan::DestroyFramebuffer()
 
 bool GPU_HW_Vulkan::CreateVertexBuffer()
 {
-  return m_vertex_stream_buffer.Create(VK_BUFFER_USAGE_VERTEX_BUFFER_BIT, VERTEX_BUFFER_SIZE);
+  const bool result = m_vertex_stream_buffer.Create(VK_BUFFER_USAGE_VERTEX_BUFFER_BIT, VERTEX_BUFFER_SIZE);
```
stenzek, line 851:
```
if (!m_vertex_stream_buffer.Create(VK_BUFFER_USAGE_VERTEX_BUFFER_BIT, VERTEX_BUFFER_SIZE))
  return false;
```

(don't call SetObjectName on a null object)

src/core/gpu_hw_vulkan.cpp
```diff
@@ -1723,12 +1838,22 @@ bool GPU_HW_Vulkan::CreateTextureReplacementStreamBuffer()
     return false;
   }
 
+  Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), m_texture_replacment_stream_buffer.GetBuffer(),
+                              "Texture Replacement Stream Buffer");
+  Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(),
+
```
stenzek, line 1844: Extra newline

src/frontend-common/vulkan_host_display.cpp
```diff
@@ -1026,6 +1047,14 @@ bool VulkanHostDisplay::CheckPostProcessingRenderTargets(u32 target_width, u32 t
     {
       return false;
     }
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), m_post_processing_input_texture.GetImage(),
+                                "Post Processing Input Texture");
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), m_post_processing_input_texture.GetView(),
+                                "Post Processing Input Texture View");
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(),
+
```
stenzek, line 1055: Extra newline

src/frontend-common/vulkan_host_display.cpp
```diff
@@ -1047,6 +1076,18 @@ bool VulkanHostDisplay::CheckPostProcessingRenderTargets(u32 target_width, u32 t
       {
         return false;
       }
+      Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(),
+
```
stenzek, line 1080: Extra newline

src/frontend-common/vulkan_host_display.cpp
```diff
@@ -1047,6 +1076,18 @@ bool VulkanHostDisplay::CheckPostProcessingRenderTargets(u32 target_width, u32 t
...
+                                  pps.output_texture.GetImage(),
+                                  ("Post Processing Output Texture" + std::to_string(i)).c_str());
```
stenzek, line 1082: Format string instead of std::string construction/concatenation

src/common/vulkan/context.cpp
```diff
@@ -644,6 +645,8 @@ bool Context::CreateCommandBuffers()
       LOG_VULKAN_ERROR(res, "vkCreateCommandPool failed: ");
       return false;
     }
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), resources.command_pool,
+                                ("Frame Command Pool " + std::to_string(frame_index)).c_str());
```
stenzek, line 649: Still using format string here

src/common/vulkan/context.cpp
```diff
@@ -654,6 +657,8 @@ bool Context::CreateCommandBuffers()
       LOG_VULKAN_ERROR(res, "vkAllocateCommandBuffers failed: ");
       return false;
     }
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), resources.command_buffer,
+                                ("Frame Command Buffer " + std::to_string(frame_index)).c_str());
```
stenzek, line 661: And here

src/common/vulkan/context.cpp
```diff
@@ -663,7 +668,8 @@ bool Context::CreateCommandBuffers()
       LOG_VULKAN_ERROR(res, "vkCreateFence failed: ");
       return false;
     }
-
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), resources.fence,
+                                ("Frame Fence" + std::to_string(frame_index)).c_str());
```
stenzek, line 672: Same here

src/common/vulkan/context.cpp
```diff
@@ -683,6 +689,10 @@ bool Context::CreateCommandBuffers()
       LOG_VULKAN_ERROR(res, "vkCreateDescriptorPool failed: ");
       return false;
     }
+    Vulkan::Util::SetObjectName(g_vulkan_context->GetDevice(), resources.descriptor_pool,
+                                ("Frame Descriptor Pool" + std::to_string(frame_index)).c_str());
```
stenzek, line 693: Same here

PR #2440, opened 2021-07-24
CI: Generate a self signed key for UWP signing pull requests

stenzek:


stenzek:
Not relevant since I'm not shipping UWP anymore.

Issue #2179, opened 2021-05-26
Debugger Improvements and Cheat Manager Improvements


niko1point0:
In the CPU Debugger's "Memory" window, can there be a window to type an address and immediately view that memory? I saw the Disassembly window had "CTRL + G", so could it be possible to click on Debugger window or Memory window to toggle one as active, and then press CTRL + G for the window?

Also in the "Memory" window, can it be possible to type new bytes manually, or copy/paste a bunch of hex from HxD into RAM? Previously I've done this in Cheat Engine, but I'll remove the need to find base address of PSX memory in Cheat Engine if I can do this in the "Memory" window of the CPU Debugger.

In the Cheat Manager's Memory Scanner, can there be a feature for adding pointers when clicking "Add Manual Address", similar to how Cheat Engine lets you add pointers?

Also in Cheat Manager, can there be a feature to type equations into the address? For example, rather than always typing "8008d2ac", would it be possible to type "80080000 + d2ac"? That would make it easier to quickly handle known offsets to pointers while reverse engineering.

Thank you for all the hard work on Duck, I love the emu so much, and you are highly appreciated!

DCxDemo:
meanwhile, it also would be useful to know the root address of PSX RAM for memory manipulation via external programs (i.e. cheat engine) without the need to lookup for some game data first, like just print the base pointer somewhere. or maybe even add an option to allocate that to a constant address, if possible.

PugsyMAME:
> In the CPU Debugger's "Memory" window, can there be a window to type an address and immediately view that memory? I saw the Disassembly window had "CTRL + G", so could it be possible to click on Debugger window or Memory window to toggle one as active, and then press CTRL + G for the window?

CTRL+D lets you set the address in the debugger Memory window.

PugsyMAME:
I've made a modification to the Cheat Memory Scanner to show the virtual 128bit address of where the Duckstation PS1 RAM lives. And you can view the address in HxD but I don't know if it's possible to change the actual data (I can edit it HXD but it doesn't seem to alter in Duckstation), might be useful for copying stuff out and that's about it. 

Haven't raised a PR for it yet, don't know if it's that useful? 

![image](https://user-images.githubusercontent.com/18102600/128637658-6d8df738-0ce8-4719-979b-6a88b37416f2.png)

olsongleb:
please make it possible to set a breakpoint for reading memory

COWCATGames:
[...]

PR #2473, opened 2021-08-03
Debugger: Add conditional data breakpoints, and a few minor bug fixes

mateusfavarin:
I've implemented conditional data breakpoints for variable memory sizes. Duckstation was storing the address of each breakpoint as the information to check for when an execution breakpoint is valid. I expanded this idea by creating the struct `DebugAddress`, which encapsulates an `address`, a `size` and a `debug_mode`. The user can create a `DebugAddress` by adding a breakpoint using the UI, and select multiple conditions for the breakpoint: read, write, changed value or execution.

The execution breakpoint that was implemented on duck had to be modified by adding an extra check for the `debug_mode`, so only execution breakpoints can trigger that event. The data breakpoint was implemented by adding a function that looks ahead of what the next CPU instruction is going to be, and if the next instruction matches the conditions of any `DebugAddress`, then the emulation is paused before the CPU executes that instruction. The UI had to be changed in order for the user to create such breakpoints.

A few minor bug fixes unrelated to that:

- Change breakpoint list ID to reset at 1, just like it's constructed.
- When a breakpoint is removed, decrement the ID of every other breakpoint with an ID greater than the one that were removed.
- When breakpoints are cleared, update the breakpoint list UI to give a visual feedback to the user.
- Fix the breakpoint not disabling when you toggle the checkbox in the breakpoint list.
- Change instruction format to always display hexadecimal values for consistency, and also change the `sll zero, zero` to `nop`.
- Add the possibility to edit and delete breakpoints by interacting with the mouse in the breakpoint list.

This PR addresses some of the issues mentioned at #2179.


stenzek:
Thanks! I'll get around to reviewing this soon, super busy with work this week :(

src/core/cpu_core.cpp
```diff
@@ -627,6 +690,126 @@ void DisassembleAndPrint(u32 addr, u32 instructions_before /* = 0 */, u32 instru
   }
 }
 
+ALWAYS_INLINE_RELEASE static bool ConditionalBreakpointLookAhead(u32 inst_bits)
+{
+  const u32 pc = g_state.regs.pc;
+  Instruction inst = g_state.current_instruction;
+  inst.bits = inst_bits;
```
stenzek, line 697: Redundant assignment

src/core/cpu_core.cpp
```diff
@@ -627,6 +690,126 @@ void DisassembleAndPrint(u32 addr, u32 instructions_before /* = 0 */, u32 instru
...
+
+  if (System::IsPaused() || pc == s_last_data_breakpoint_check_pc)
+    return false;
+
+  // Skip nops. Makes PGXP-CPU quicker, but also the regular interpreter.
+  if (inst.bits == 0)
+    return false;
+
+  switch (inst.op)
```
stenzek, line 706: Just read `g_state.current_instruction`

src/core/cpu_core.cpp
```diff
@@ -1893,11 +2115,11 @@ static void ExecuteImpl()
       {
         Cop0ExecutionBreakpointCheck();
 
-        if (BreakpointCheck())
-        {
-          // continue is measurably faster than break on msvc for some reason
-          continue;
-        }
+        if (ExecutionBreakpointCheck())
```
stenzek, line 2118: `if (ExecutionBreakpointCheck() || ConditionalBreakpointLookAhead())`

src/core/cpu_core.cpp
```diff
@@ -627,6 +690,126 @@ void DisassembleAndPrint(u32 addr, u32 instructions_before /* = 0 */, u32 instru
...
+    {
+      const VirtualMemoryAddress addr = ReadReg(inst.i.rs) + inst.i.imm_sext32();
+      DataBreakpointCheck<MemoryAccessType::Read>(addr);
+      DataBreakpointCheck<MemoryAccessType::Read>(addr + 1);
+    }
+    break;
+    case InstructionOp::lw:
+    {
+      const VirtualMemoryAddress addr = ReadReg(inst.i.rs) + inst.i.imm_sext32();
+      DataBreakpointCheck<MemoryAccessType::Read>(addr);
```
stenzek, line 726: This is going to get *very* slow. It would probably be better to align the test address down, and do a single range check in `DataBreakpointCheck`.

src/core/types.h
```diff
@@ -17,6 +17,21 @@ enum class MemoryAccessSize : u32
   Word
 };
 
+struct DebugAddress
```
stenzek, line 20: Rather than combining the breakpoints, it'd be better to store the execution and data breakpoints in separate arrays (fewer loops/comparisons for every guest instruction).

stenzek (commented):
Only looked at the CPU side for now, I'm a little concerned about the performance impact given the fact that this code can potentially execute up to 33.8 million times per second.

Issue #1662, opened 2021-02-21
Opening cheat manager and pressing "New Search" eventually crash DS


Sekai9:
**Game details**
SLUS-00748 - Resident Evil 2: Dual Shock Ver. (Disc 1) [Leon]
or
SLUS-00756 - Resident Evil 2: Dual Shock Ver. (Disc 2) [Claire]
^ Redump verified.

**Description of the issue / bug**
DuckStation crashes after you perform the steps below and hit Power Off.

**Steps to reproduce**
First, disable "Save state on exit."
1> Load RE2 (Claire disc seems to trigger this more easily)
2> Open the Cheat Manager, go to Memory Scanner, press New search, hit ok, close the window.
3> Press Power Off, DS should crash.

Seems pretty random, so it might take a few tries until it crashes. Just repeat all the steps.
Both Interpreter and Recompiler are affected.

**Enhancements information**
None, it's a clean portable 'install'.

**Hardware/software**
Intel i5-5300U + Intel HD 5500, Windows 10.

**Emulator version**
0.1-3250-g9d05aa21
[crash.zip](https://github.com/stenzek/duckstation/files/6017155/crash.zip)

stenzek:
Memory scanner is persistent across invocations as of https://github.com/stenzek/duckstation/commit/4598bc789b5630dc819eeb7863ce62357dcfc17c.

PR #2485, opened 2021-08-07
Possible fix for bug mentioned in  Discord

PugsyMAME:
May possibly  be a fix for this as well:- https://github.com/stenzek/duckstation/issues/1662


Nucleoprotein:
Maybe this prevents crash but... why Bus::g_ram is null there?

PugsyMAME:
It crashes when it goes back to the frontend, after the power down of the emulator core. I'm surmissing that because the memory scanner window was closed down prior to power down it didn't close it properly when you hit power down. And it possibly does a clean up after and Bus::g_ram would be null at that time.

But I'm just guessing - it was a quick fix and didn't investigate more.

andercard0:
> 
> 
> May possibly be a fix for this as well:- #1662

Yes, boot and debug fixed!

image

Nucleoprotein:
@PugsyMAME 
Thanks for exploitation, seems good.

stenzek:
There's still a race condition here - Bus::g_ram could go null between the if and memcpy, since the system shutdown happens on the emu thread. I've held off in fixing this because the proper solutions aren't really great (make shutdown synchronous, or do the search on the emu thread, the latter is probably the least rubbish of the two).

stenzek:
RAM stays persistently mapped across system instances now, so the original problem should no longer be an issue.

F6 type was already added in https://github.com/stenzek/duckstation/commit/0fd593eb426b4982413de02b39e26c7d43d7807e.

PR #2625, opened 2021-09-25
generate a wayland appimage

qurious-pixel:
- update patchelf to 0.12
- integrate qt5-wayland plugins
- check for wayland display environment

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -242,7 +242,7 @@ jobs:
       shell: bash
       run: |
         sudo apt-get update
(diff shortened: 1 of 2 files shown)


stenzek:
Not sure why this has to be a separate release file? Can't x11 and wayland share the same appimage?

The other (perhaps separate issue) here is Qt6 - no idea how to handle that yet. Windows has been on it since 6.1 was released, but Linux is stuck on 5.12 or whatever's in Ubuntu 20.04.

qurious-pixel:
The `duckstation-wayland-x64.AppImage` is both x11 and wayland. I separated it for those that specifically want wayland, the logic picks wayland environment over x11, but if not found, falls back.  Qt6 is almost at feature parity with Qt5, but not quite there yet.

<details>
<summary> Assets </summary>

image

</details>

stenzek:
It actually depends on your desktop environment. On kwin it uses Wayland by default, on gnome it uses X (because it sucks and has no server side decorations).

I still don't see why a second release is needed.

PR #2629, opened 2021-09-28
Update imgui to v1.85 (WIP)

Wunkolo:
Updates imgui to the current WIP of v1.85. Updates all of the
OpenGL/DX11/DX12/Vulkan backends to use the latest Imgui patterns.

This also fixes a Vulkan issue in particular regarding mapped memory not
being a multiple of `VkPhysicalDeviceLimits::nonCoherentAtomSize`.

```
[    8.5543] E(DebugMessengerCallback): Vulkan debug report: (VUID-VkMappedMemoryRange-size-01389) Validation Error: [ VUID-VkMappedMemoryRange-size-01389 ] Object 0: handle = 0xc84ae90000000996, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xee4872d | vkFlushMappedMemoryRanges: Size in pMemRanges[0] is VK_WHOLE_SIZE and the mapping end (0x9df8 = 0x0 + 0x9df8) not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (0x40) and not equal to the end of the memory object (0x9e00). The Vulkan spec states: If size is equal to VK_WHOLE_SIZE, the end of the current mapping of memory must either be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize bytes from the beginning of the memory object, or be equal to the end of the memory object (https://vulkan.lunarg.com/doc/view/1.2.176.1/windows/1.2-extensions/vkspec.html#VUID-VkMappedMemoryRange-size-01389)
```

### DIFF
--- a/dep/imgui/CHANGELOG.txt
+++ b/dep/imgui/CHANGELOG.txt
@@ -3,16 +3,18 @@ CHANGELOG
 
 This document holds the user-facing changelog that we also use in release notes.
 We generally fold multiple commits pertaining to the same topic as a single entry.
-Changes to the examples/bindings are included within the individual .cpp files in the examples/ folder.
+Changes to backends are also included within the individual .cpp files of each backend.
 
-RELEASE NOTES:                  https://github.com/ocornut/imgui/releases
-REPORT ISSUES, ASK QUESTIONS:   https://github.com/ocornut/imgui/issues
-COMMITS HISTORY:                https://github.com/ocornut/imgui/commits/master
+RELEASE NOTES:          https://github.com/ocornut/imgui/releases
+REPORT ISSUES:          https://github.com/ocornut/imgui/issues
+DISCUSS, ASK QUESTIONS: https://github.com/ocornut/imgui/discussions
+FAQ                     https://www.dearimgui.org/faq/
+WIKI                    https://github.com/ocornut/imgui/wiki
 
 WHEN TO UPDATE?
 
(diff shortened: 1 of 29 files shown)


Wunkolo:
Currently facing an issue involving `HostDisplayTexture`s causing things like the loading screen(in relevant APIs that use the "Compiling Pipelines" loading screen) to bug out _sometimes_ and it would show the font-atlas or just crash on DX12 at `ctx->SetGraphicsRootDescriptorTable(1, texture_handle);`. Otherwise, it's pretty stable so far and fixes the Vulkan issue, [among many other things](https://github.com/ocornut/imgui/blob/master/docs/CHANGELOG.txt).
image

This might have to do with the fact that we should be calling functions like `ImGui_ImplDX12_NewFrame();` before calling `ImGui::NewFrame();` or some other stuff involving the coherency of our image handles.

stenzek:
There are modifications to the imgui code in tree, I know controller navigation was one of them. So I'll have to go through and find all those changes.

Wunkolo:
If the cost of updating all of Imgui, just to fix this one Vulkan bug in particular, is too much then this PR might be worth closing if we can just possibly recreate the upstream fix by editing the files ourselves. Unless we have been meaning to update it anyways.

Wunkolo:
Superseded by d34477a21447364369604566a4d97126f5e626fe

PR #2636, opened 2021-10-01
Fix 4k on series consoles

tunip3:
Remove check for original Xbox One as it doesn't work on series consoles as they would be detected as original model Xboxes meaning that 4k wouldn't work.
Getting rid of this check has no adverse effects on the original Xbox one and makes 4k work properly on series x.
@gamr13 has validated that 4k now works on Xbox series consoles and I have validated that there are no adverse effects to original model Xbox ones.
image


stenzek:
Interesting that it reports as an X1 despite the other enum values, and an X1S reports itself..

PR #2682, opened 2021-10-27
Qt6 with Qt5 fallback

qurious-pixel:


j8r:
Another option is to use Alpine to statically link the program, Qt6 is already available there.

Also for cross-distribution, so far the [flatpak](https://github.com/flathub/org.duckstation.DuckStation) has worked well, minus an issue related to Wayland (Flatpak was not the cause).

qurious-pixel:
Updated to use CMAKE command USE_QT6 to select which version of CMAKE to use.
Searches submodule path for Qt6
Fallback searches for Qt5 (Minimum version 5.15)
CI can use submodule or download external package

scripts/generate_appimages.sh
```diff
@@ -88,11 +88,23 @@ ${BUILD_DIR}/squashfs-root/AppRun \
 sed -i 's|exec "$this_dir"/AppRun.wrapped "$@"|exec env LD_LIBRARY_PATH="$this_dir"/usr/lib:$LD_LIBRARY_PATH "$this_dir"/AppRun.wrapped "$@"|' \
   ${BUILD_DIR}/duckstation-qt.AppDir/AppRun
 
+if [[ $(qmake -v | grep 'Qt' | awk ' { print $4 } ' | cut -d '.' -f 1) > 5 ]]; then
```
j8r, line 91:
```suggestion
if [ $(qmake -v | awk ' END { print $4 } ' | cut -d '.' -f 1) -gt 5 ]; then
```
More compact and more standard syntax.

stenzek:
Redundant as DuckStation is qt6-only now.

PR #2711, opened 2021-11-07
Fix usage of `free` for stack variable

Dushistov:
Based on warning from pvs static analyzer (https://habr.com/ru/company/pvs-studio/blog/586700/)


stenzek:
Nice catch. I guess this is very unlikely to trip so that's why it wasn't noticed :)

PR #2712, opened 2021-11-07
AnalogController: Fix legacy rumble conditions and implement status byte toggling

ggrtk:
Just some really old commits from hardware testing.


stenzek:
Great, many thanks!

PR #2735, opened 2021-11-17
Update chtdb.txt

Risae:
Adds the following cheats to Gran Turismo 1 & 2:

Gran Turismo 1:
- 60 FPS (NTSC-U 1.1, NTSC-U 1.0, PAL, JP NTSC-J)
- Simulation timescale in Arcade (NTSC-U 1.1, PAL)

Gran Turismo 2:
- 16:9 Widescreen Hack 2.0 (NTSC-U 1.2, NTSC-U 1.1, PAL, NTSC-J 1.1, NTSC-J 1.0)
- 21:9 Widescreen Hack 2.0 (NTSC-U 1.2, NTSC-U 1.1, PAL, NTSC-J 1.1, NTSC-J 1.0)
- 60 FPS (NTSC-U 1.2, NTSC-U 1.1, PAL, Arcade Disc: JP NTSC-J 1.0, Simulation Disc: JP NTSC-J 1.1, NTSC-J 1.0)
- Metric units code (NTSC-U 1.2, NTSC-U 1.1, PAL)
- Full detail AI cars (NTSC-U 1.1, NTSC-U 1.2, PAL, NTSC-J 1.1, NTSC-J 1.0)
- Slightly higher draw distance (NTSC-U 1.2, NTSC-U 1.1, PAL, NTSC-J 1.1, NTSC-J 1.0)
- Use 8MB RAM for polygon buffers (NTSC-U 1.1, NTSC-U 1.2, PAL, NTSC-J 1.1, JP NTSC-J 1.0)

Credits go to cookieplmonster
https://cookieplmonster.github.io/mods/gran-turismo/
https://cookieplmonster.github.io/mods/gran-turismo-2/

### DIFF
--- a/data/database/chtdb.txt
+++ b/data/database/chtdb.txt
@@ -113590,7 +113590,6 @@ A0010A4C 00008021
 90010A4C 8C70006C
 00000000 FFFF
 00000000 FFFF
-
 #True Endurance tweak (NTSC-J 1.1)
 ;"Help" the game unload segments that normally are left unwritten to,
 ;so further cheat detection is more reliable. This is safe as it overwrites dead code.
--- a/data/database/chtdb.txt
+++ b/data/database/chtdb.txt
@@ -28034,6 +28034,92 @@ A7016A7A 02250000
 D121F88A AEB4
 A7016A88 80000000
 00000000 FFFF
+#L3 to toggle Mirror (tap) and HUD (hold)
+A403EC74 02602021
+D7010001 00000200
+A0029524 1040000C
+;Always on
+90029524 00000001
+A0029524 0800A556
+;Default
+90029524 1040000C
+A0029524 00000000
+;Always off
+90029524 0800A556
+;Fixup canary
+A0029524 00000001
+90029524 00000000
+00000000 FFFF
+00000000 FFFF
+A403EC74 02602021
+D701003C 00000200
+F5029434 0022A530
+F5029436 14400800
+F5029424 BA060000
+F5029426 0C000000
+00000000 FFFF
+00000000 FFFF
+#Enable extra cameras in race
+A401F888 AEB40008
+A701031C 00030009
+A7010370 40F045C1
+;Replay off
+C30A92BC 0001
+A701171C 0106010E
+A7011778 45E945D7
+00000000 FFFF
+00000000 FFFF
+A401F888 AEB40008
+;Hold R1 to trigger a cinematic camera
+;Replay off
+C30A92BC 0001
+D701001E 01000008
+F5010148 40B6427F
+F5010A4C 006C8021
+F5010A4E 8C700000
+301FFA89 0002
+00000000 FFFF
+00000000 FFFF
+00000000 FFFF
+;Restore everything when replay is enabled
+A401F888 AEB40008
+C40A92BC 0000
+D001171C 010E
+8001171C 0106
+D0011778 45D7
+80011778 45E9
+A0010A4C 00008021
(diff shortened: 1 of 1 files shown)


stenzek:
@CookiePLMonster are you okay with these being included?

CookiePLMonster:
Looks good, thanks! Just two requests from me:

1. GT2 already ships two other 16:9 widescreen cheats, might be a wise idea to delete them as else Duck will ship 3 different cheats.
2. What's the deal with NTSC-J versions? Looks like they are merged, but then cheats are separate. Might be a good idea to split NTSC-J 1.0 and 1.1.

Risae:
The Man himself! Thank You for creating those updated cheats. They are, truly, awesome!

> Looks good, thanks! Just two requests from me:
> 
>1. GT2 already ships two other 16:9 widescreen cheats, might be a wise idea to delete them as else Duck will ship 3 different cheats.

I titled those "2.0" in order to differentiate them from the originals. If you guys think its better to remove the old ones then i'll do that. In my opinion we should keep them in case anyone wants to try them out.

>2. What's the deal with NTSC-J versions? Looks like they are merged, but then cheats are separate. Might be a good idea to split NTSC-J 1.0 and 1.1.

I was wondering about that too, since i wasn't able to find the entry for version 1.1 inside the DB.
Looking at it again i think that version 1.1 doesn't actually have any cheats in the Database?
This should be 1.0:

; [ Gran Turismo 2 (Arcade + Gran Turismo Mode) (Japan, Asia) {SCPS-10116 | SCPS-10117} ]
:SCPS-10116
:SCPS-10117

And this should be 1.1?:

; [ Gran Turismo 2 (Disc 2) (Gran Turismo) (Japan) Rev 1 {SCPS-91327} ]
http://redump.org/disc/1898/

If this is right then i will make a new entry for 1.1 below "**; [ Gran Turismo 2 (Disc 2) (Gran Turismo) (Japan) Rev 1 {SCPS-91327} ]**" and add the cheats there.

CookiePLMonster:
IMO there should be no need for shipping the other cheats as they are provably less feature complete than the one I came up with (most notably, broken rear view mirror) but I'll let @stenzek decide.

The situation with NTSC-J is annoying since Arcade JP 1.0 and Simulation JP 1.0 are different. I'll have to check the cheat db locally with those versions and I'll get back to you, but at the first glance your idea seems okay.

stenzek:
Sounds reasonable to me. @PugsyMAME did you have any opinion? (given you originally put the database together)

PugsyMAME:
The Gran Turismo 1 & 2 cheats need a lot of work and testing done anyway, so crack on I say. Plus any work that I don't have to do is good news..... I've been avoiding touching Gran Turismo 1+2 cheats for a while as frankly they are a mess which will require LOTs of cleaning up+testing.

stenzek:
Okay great, after you've made those changes I'll merge.

Risae:
@CookiePLMonster @stenzek added the changes

CookiePLMonster:
Untested, but it looks good. I'll give it a spin later today and edit the post.

shikulja:
@CookiePLMonster
Does not relate to the topic, if possible, in the future, is it possible to make such cheats for the road rash series?

CookiePLMonster (approved):
Tested, works fine with one caveat - because all game revision share name/serial, cheat DB only auto-loads cheats for the most recent game revision and not the one you are actually playing. 

If this is acceptable for @stenzek I think it can be merged.

PR #2752, opened 2021-11-28
Make appimage respect "QT_QPA_PLATFORM" env var.

BailShamber:
This is so users who want to opt out from using wayland by using `QT_QPA_PLATFORM=xcb` don't get confused.


BailShamber:
This needs a change to make it safer if "$WAYLAND_DISPLAY" doesn't return a valid wayland display.

PR #2784, opened 2021-12-26
Use Ninja for macOS builds on GitHub Actions

Calinou:
Ninja makes for faster builds compared to Make.


stenzek:
Won't make any difference since the runners only have 2C/T, but sure

Calinou:
> Won't make any difference since the runners only have 2C/T, but sure

According to [this page](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), macOS instances have 3-core CPUs now (and twice the amount of RAM compared to Windows/Linux runners).

PR #2804, opened 2022-01-17
M3U: Preloading to RAM for single-disc playlists.

brianblakely:
HasSubImages will return false when the playlist has only a single sub-image.

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -655,7 +655,7 @@ std::unique_ptr<CDImage> OpenCDImage(const char* path, Common::Error* error, boo
 
   if (force_preload || g_settings.cdrom_load_image_to_ram)
   {
-    if (media->HasSubImages())
+    if (media->HasSubImages() && media->GetSubImageCount() > 1)
     {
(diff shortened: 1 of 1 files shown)


stenzek:
Not sure what the advantage would be to this, other than making the files incorrectly show in the game list as discs instead of playlists (which is what they are)?

brianblakely:
> Not sure what the advantage would be to this, other than making the files incorrectly show in the game list as discs instead of playlists (which is what they are)?

For instance, discs loaded from M3U in this fashion will now preload into RAM when that setting is enabled.

stenzek:
Did you clear the cache/rescan all games? The type is cached alongside everything else.

> For instance, discs loaded from M3U in this fashion will now preload into RAM when that setting is enabled.

If this is the only reason, it would be better to do a subimage count check in the preload routine instead, IMO.

brianblakely:
> If this is the only reason, it would be better to do a subimage count check in the preload routine instead, IMO.

Thank you, PR is amended to reflect this.

PR #2808, opened 2022-01-23
Add CHD precache support

realnc:
This uses the chd_precache() function of libchdr which just loads the
CHD image to memory without decompressing it. So this is "weaker" than
the full preload to RAM option, which fully decompresses the CHD image,
but it still solves the problem of having the CHD images stored on
unreliable network storage.

The point of this is to be able to use M3U files to run multi-disc games
and still be able to work around unreliable storage. The full preload to
RAM option does not work with M3U files. This option does.

CHD precache is ignored (and the option disabled in the UI) if the full
preload to RAM option is enabled.

### DIFF
--- a/src/common/cd_image.cpp
+++ b/src/common/cd_image.cpp
@@ -6,7 +6,7 @@
 #include <array>
 Log_SetChannel(CDImage);
 
-CDImage::CDImage() = default;
+CDImage::CDImage(OpenFlags open_flags) : m_open_flags(open_flags) {}
 
 CDImage::~CDImage() = default;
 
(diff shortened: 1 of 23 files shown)


src/common/cd_image.cpp
```diff
@@ -47,7 +48,8 @@ std::unique_ptr<CDImage> CDImage::Open(const char* filename, Common::Error* erro
   }
   else if (StringUtil::Strcasecmp(extension, ".chd") == 0)
   {
-    return OpenCHDImage(filename, error);
+    const bool precache = g_settings.cdrom_precache_chd && !g_settings.cdrom_load_image_to_ram;
```
stenzek, line 51: Common code should not depend on core. The easiest way around this would probably be to add a precache parameter, or better yet, a flags parameter.

realnc: Is there an existing flags enum somewhere I can extend with a precache field?

stenzek:
Closed by author request.

PR #2822, opened 2022-04-11

zkdpower:


zkdpower:
OSD "no such file or directory" is no string for translating. please add it.

PR #2836, opened 2022-07-23
housekeeping

andercard0:
Removed some commentaries from PCSX2 in the code


stenzek:
Will do this one locally with a more detailed commit message.

PR #2841, opened 2022-07-24
Update Simple-Chinese language file to latest？

zkdpower:
Update Simple-Chinese language file to latest？seems string of "option tip" are waiting you complementing.


stenzek:
Sorry, I don't accept .qm files, iirc they're version specific (and everything except windows is on qt6.3).

zkdpower:
oh. i forget the format.

PR #2842, opened 2022-07-25
Little improvements to the post-processing tab

Iniquitatis:
It was a bit hard to read the shader parameter names when there were no margins.

---
### Before
Before

### After
After


zkdpower:
perhaps the window's position of "Tweak/Hacks" also needs some adjusting. currently is too high, too close to top.

stenzek:
Sounds reasonable, thanks!

zkdpower:
do you find a bug, the interface of post-processing will suddenly jump in the middle when you open&close some filter.

PR #2847, opened 2022-07-26
Fix building if WITH_CHEEVOS is not defined.

MaddTheSane:


stenzek:
Closing as I'm not planning on supporting compiling without achievements in the future, it makes no sense to disable critical functionality like this.

PR #2849, opened 2022-07-26
Fix Thread Naming on macOS

MaddTheSane:
This fixes thread naming on macOS, where the code looks like it was targeted for FreeBSD or OpenBSD instead of Darwin/macOS (macOS doesn't have a `<pthread_np.h>` header. While it does have `pthread_setname_np`, it is _not_ `pthread_set_name_np`.).

### DIFF
--- a/src/common/threading.cpp
+++ b/src/common/threading.cpp
@@ -30,7 +30,7 @@
 #include <mach/mach_time.h>
 #include <mach/semaphore.h>
 #include <mach/task.h>
-#include <pthread_np.h>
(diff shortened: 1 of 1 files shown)


stenzek:
This is redundant with https://github.com/stenzek/duckstation/commit/0006c54c46c1ea87809adae0d027e3eb5045c25e#diff-d5774961cd0ccb7e3e06b63ff439c5c5ab1c2e07fa031cb29c5c94825118c7e9

PR #2860, opened 2022-08-05
fix throttler and vsync related issues

Guwange:
This commit may fix issues like below:

  * VSYNC will be disabled after pause (including disc change, etc.)
  * 'Unlimited' setting for fast-forward / turbo not working. (And it seems other speed setting also not working..)

This commit fixes below functions:

  * System::UpdateSpeedLimiterState()
    * m_throttler_enabled should be set if target_speed != 1.0f, not 0.0f. 
      - m_throttler_enabled was set to true even in case emulator is running at 100%.
      - m_throttler_enabled was set to false in case fast_forward/turbo was enabled but set to 'Unlimited'
    * ShouldUseVSync() should be called after s_target_speed is updated with target_speed.
      - ShouldUseVSync() uses IsRunningAtNonStandardSpeed() internally, and it refers s_target_speed. 

  * System::ShouldUseVSync()
    * I guess this function should return 'true' in case 'g_settings.video_sync_enabled == true' && 'm_throttler_enabled == false'. but current implementation returns false.

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -2210,13 +2210,6 @@ void System::UpdateSpeedLimiterState()
     }
   }
 
-  const bool video_sync_enabled = ShouldUseVSync();
-  const float max_display_fps = (!IsRunning() || m_throttler_enabled) ? 0.0f : g_settings.display_max_fps;
-  Log_InfoPrintf("Target speed: %f%%", target_speed * 100.0f);
-  Log_InfoPrintf("Using vsync: %s", video_sync_enabled ? "YES" : "NO");
-  Log_InfoPrintf("Max display fps: %f (%s)", max_display_fps,
-                 m_display_all_frames ? "displaying all frames" : "skipping displaying frames when needed");
-
   if (IsValid())
   {
     AudioStream* stream = g_spu.GetOutputStream();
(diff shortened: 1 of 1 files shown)


src/core/system.cpp
```diff
@@ -2191,7 +2191,7 @@ void System::UpdateSpeedLimiterState()
   float target_speed = m_turbo_enabled ?
                          g_settings.turbo_speed :
                          (m_fast_forward_enabled ? g_settings.fast_forward_speed : g_settings.emulation_speed);
-  m_throttler_enabled = (target_speed != 0.0f);
+  m_throttler_enabled = (target_speed != 1.0f);
```
stenzek, line 2194: This logic doesn't sound correct. 1.0 is "normal" speed, where you want to throttle.

Guwange: Oh I see.. 

Then I think the m_throttler_enabled related interdependency between UpdateSpeedLimiterState() and ShouldUseVSync() should be removed.
(m_throttler_enabled will be finally set as 'false' at the end of  UpdateSpeedLimiterState() in case target_speed == 1.0f and ShouldUseVSync() return true.
But currently ShouldUseVSync() returns false in case m_throttler_enabled is 'true' and it is running at normal speed. So at the timing UpdateSpeedLimiterState() calls ShouldUseVSync(), it always return false and VSync will be never enabled.)

I think there's no need to check m_throttler_enabled in ShouldUseVSync() so I will simply remove it.

Guwange:
I updated the commit and force pushed. 
Could you review it again ?  Thanks.

stenzek:
Might want to re-check this on the latest master, I ended up redoing part of this logic anyway.

Guwange:
> Might want to re-check this on the latest master, I ended up redoing part of this logic anyway ([54c2447](https://github.com/stenzek/duckstation/commit/54c2447ff3b66b8636748ee0b68c1c0c50304b87)).

I checked the latest master. 
It looks fine and the built binary also seems working fine.
I think now this PR can be closed. Thanks !

PR #2867, opened 2022-08-13
fix AchievementSummary/Notification font issue for ja/zh-cn

Guwange:
This PR reverts imgui_manager.cpp modification made by below commit:

------------
commit 853e9a8a6a7234caeaa2ca75544375318cc86d68
Author: Connor McLaughlin <stenzek@gmail.com>
Date:   Fri Jul 22 21:51:00 2022 +1000

    Qt: Fix updater
------------

AchievementSummary (and Leaderboard submittion) uses Notification, and Notification is drawn using medium_font and large_font.
There are already some translations for AchievementSummary/Leaderboard.
This PR fixes issue that these translated messages are currently shown with ???? string.

### DIFF
--- a/src/frontend-common/imgui_manager.cpp
+++ b/src/frontend-common/imgui_manager.cpp
@@ -30,7 +30,7 @@ static void SetStyle();
 static void SetKeyMap();
 static bool LoadFontData();
 static bool AddImGuiFonts(bool fullscreen_fonts);
-static ImFont* AddTextFont(float size, bool full_range);
+static ImFont* AddTextFont(float size);
 static ImFont* AddFixedFont(float size);
 static bool AddIconFonts(float size);
 static void AcquirePendingOSDMessages();
@@ -419,7 +419,7 @@ bool ImGuiManager::LoadFontData()
   return true;
 }
 
-ImFont* ImGuiManager::AddTextFont(float size, bool full_range)
+ImFont* ImGuiManager::AddTextFont(float size)
 {
   static const ImWchar default_ranges[] = {
     // Basic Latin + Latin Supplement + Central European diacritics
@@ -445,7 +445,7 @@ ImFont* ImGuiManager::AddTextFont(float size, bool full_range)
   cfg.FontDataOwnedByAtlas = false;
   return ImGui::GetIO().Fonts->AddFontFromMemoryTTF(s_standard_font_data.data(),
                                                     static_cast<int>(s_standard_font_data.size()), size, &cfg,
-                                                    (s_font_range && full_range) ? s_font_range : default_ranges);
+                                                    s_font_range ? s_font_range : default_ranges);
 }
 
 ImFont* ImGuiManager::AddFixedFont(float size)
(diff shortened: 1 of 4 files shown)


Guwange:
I also added very minor fix for OSD message about Challenge Mode.

src/frontend-common/imgui_manager.cpp
```diff
@@ -488,14 +488,13 @@ bool ImGuiManager::AddImGuiFonts(bool fullscreen_fonts)
 
   if (fullscreen_fonts)
   {
-    // Don't add full range for FSUI, we probably don't have the texture space, and it's not translated yet anyway.
```
stenzek, line 491: The comment here illustrates the issue, especially at 4K display resolutions, the font texture gets way too large to include all font ranges. Hence why only the normal-sized font (used for OSD messages) uses the full range.

Guwange: Thanks for your comment.
I see what is the problem... so for ja/zh-cn, is it better not to add translation for these Achievement Summary / Notifications ..? (to avoid showing ???? and instead to show original English messages)

If I could generate custom Glyph Ranges based on the translation text, using AddText() of the ImFontGlyphRangesBuilder  like below,
https://github.com/ocornut/imgui/blob/master/docs/FONTS.md#using-custom-glyph-ranges
I guess we can solve that large texture issue, but unfortunately current QTranlator seems not providing functions to retrieve translated texts (messages).

How do you think if I create small offline tool which output ImVector<ImWchar> ranges based on *.ts files, and embed that output around below code in qthost.cpp: InstallTranslator() ?

`
  if (language == QStringLiteral("ja"))
  {
    ImGuiManager::SetFontPath(GetFontPath("msgothic.ttc"));
    ImGuiManager::SetFontRange(ImGui::GetIO().Fonts->GetGlyphRangesJapanese());
  }
`

Guwange:
> How do you think if I create small offline tool which outputs ImVector<ImWchar> ranges based on *.ts files, and I manually embed that output around below code in qthost.cpp: InstallTranslator() ?

I have implemented above my idea as 3rd commit ([generate GlyphRanges for ja/zh-cn based on *.ts files](https://github.com/stenzek/duckstation/pull/2867/commits/0d03aa2b45c8fd81002193ab2962ec27f4261c0c)).
(Instead of generating ImWchar[], the script generates character list used in *.ts file, and qthost.cpp generates (converts) this list into ImWchar[].)

I think actually we don't have to include all of characters used in *.ts file - we only need characters used in OSD messages and some messages related to achievement, but currently this script will include all of characters in *.ts file.

By this modification, I guess the number of characters (glyphs) will be shrinked from about 3000 to 700 in Japanese, and from about 2500 to 900 in Chinese. I confirmed that x9 scale (for 4K) setting works in my environment (Intel GPU and nVidia GPU).

But I also think this method is somewhat tricky, and would like to hear comment from @zkdpower as well.

stenzek:
Sounds good, thanks! They should be in a separate file, because otherwise clang-format's going to make a mess of it, but I can move that.

PR #2869, opened 2022-08-16
Use existing desktop file for AppImage

j8r:
By looking at the build scripts I noticed that the existing desktop file can be used when building the AppImage.
The field changes are:
- `Name` from `DuckStation` to `DuckStation Qt`
- `Comment` from `Fast PlayStation 1 emulator` to `Fast-ish PlayStation 1 emulator`.

Other than that both files are the same.

### DIFF
--- a/scripts/make-appimage.sh
+++ b/scripts/make-appimage.sh
@@ -275,17 +275,7 @@ Plugins = ../lib/plugins
 EOF
 
 echo "Creating desktop..."
-cat > "$OUTDIR/duckstation-qt.desktop" << EOF
-[Desktop Entry]
-Type=Application
-Name=DuckStation
-GenericName=PlayStation 1 Emulator
-Comment=Fast PlayStation 1 emulator
-Icon=duckstation-qt
-TryExec=duckstation-qt
-Exec=duckstation-qt %f
-Categories=Game;Emulator;Qt;
-EOF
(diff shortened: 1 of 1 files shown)


stenzek:
My plan was to remove all those files (and the old build scripts) completely. Just forgot to do so.

j8r:
It is still better to have the desktop file as a file instead of an inline string in `make-appimage.sh`, so it can be used when we compile from source for instance.

Where to place it in the source code is another question...

j8r:
Well, you can't be bothered with issues here anymore if this is what worries you, perhaps in the discord... 🤔 

Anyhow my take on that it is totally normal not to solve issues you don't want to spend time on.
There is always this annoying issue triage to do but some projects use additional help from trusted contributors.
Others just do nothing at all and let issues be sorted out (or not).

Not a problem then, I'll create a desktop file inspired from this script.

stenzek:
I'm not against contributions. There's been a few over the time the project's been up.

What I'm not interested in, is making concessions or allowances for third party packages, which is what this was for. They're a constant source of pain (packagers breaking stuff, **especially** the AUR), and the users blame upstream instead of the packager who broke it in the first place. Hence why the AppImage is the only supported version for Linux, it works everywhere, and I can control/reproduce the build environment locally, and debug/fix any problems.

PR #2877, opened 2022-09-04
Fixed an issue about post-processing interface.

zkdpower:
Fixed an issue where the window of post-processing suddenly jumped to the center after switching filters. (special thanks to user 2048 from ppxclub.com)


stenzek:
Committed as https://github.com/stenzek/duckstation/commit/b4f2e297e92a66f51392cf57369d46c1c6ac9763 without changing the whitespace of the entire file.

PR #2887, opened 2022-09-22
update simple-chinese language file to latest.

zkdpower:


zkdpower:
Dear Author, please provide following string for foreign translation.
1."No save state found in slot #""
2."Are you sure you want to shut down the virtual machine?"

PR #2891, opened 2022-09-22
Add option to disable CRT aspect ratio accuracy correction

Superstarxalien:
DuckStation's aspect ratio stretching doesn't actually stretch the game display to match the resolution of the ratio, e.g. a game doesn't get stretched to 320x240 when the aspect ratio is set to 4:3, as it also makes the image thinner to more closely match what is displayed on CRTs, i.e. a 512px-framebuffer game set to 4:3 will be 292 pixels wide. Some games are built around the actual aspect ratio displayed on CRT screens, and for the games that aren't I added a setting for retaining the ordinary aspect ratio.

Notes:
Labelled as "CRT Pillarboxing" in settings, but it might need a better name. It also lacks a description in both Qt UI and big picture mode.
While I made sure the code worked for every aspect ratio, resolution upscale, and crop mode, two errors slipped past myself at the last minute: the display is erroneously stretched when the aspect ratio is set to 1:1 PAR, and when the "Stretch To Fill" setting is enabled, it doesn't fill the entire screen, meaning there will often be a sort of letterboxing effect.

### DIFF
--- a/src/core/gpu.cpp
+++ b/src/core/gpu.cpp
@@ -526,6 +526,9 @@ float GPU::GetDisplayAspectRatio() const
     if (m_crtc_state.display_width == 0 || m_crtc_state.display_height == 0)
       return 4.0f / 3.0f;
 
+    if (g_settings.display_crt_pillarboxing)
+      return static_cast<float>(m_crtc_state.display_vram_width) / static_cast<float>(m_crtc_state.display_height);
+
     return static_cast<float>(m_crtc_state.display_width) / static_cast<float>(m_crtc_state.display_height);
   }
   else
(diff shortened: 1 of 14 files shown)


ggrtk:
I am not entirely sure what this accomplishes that the currently available options don't provide.

> DuckStation's aspect ratio stretching doesn't actually stretch the game display to match the resolution of the ratio, e.g. a game doesn't get stretched to 320x240 when the aspect ratio is set to 4:3, as it also makes the image thinner to more closely match what is displayed on CRTs, i.e. a 512px-framebuffer game set to 4:3 will be 292 pixels wide.

The internal framebuffer dimensions, with the exception of 1:1 PAR, are not particularly relevant to the display aspect ratio. The aspect ratio setting applies only to the visible area of the CRT. You should be seeing a visible width of 292px on the Auto (Game Native) setting and a visible width of ~299px on the 4:3 setting (224 lines visible).

> Some games are built around the actual aspect ratio displayed on CRT screens, and for the games that aren't I added a setting for retaining the ordinary aspect ratio.

The "actual aspect ratio displayed on CRT screens" is sort of what the Auto (Game Native) setting accomplishes based on video digitization standards, but CRTs could have varying amounts of scaling applied depending on calibration.

stenzek:
Closing as requested by author.

PR #2918, opened 2022-10-29
Update mainwindow.cpp

xujibbs:


Nucleoprotein:
Yeah, quote is in wrong place.

stenzek:
It's fine, I changed it and force pushed over the merge commit.

PR #2922, opened 2022-11-13
Achievements: Fix building Qt frontend with achievements disabled

impiaaa:


src/duckstation-qt/mainwindow.cpp
```diff
@@ -39,10 +40,6 @@
 #include <QtWidgets/QStyleFactory>
 #include <cmath>
 
-#ifdef WITH_CHEEVOS
-#include "frontend-common/achievements.h"
```
stenzek, line 43: This won't work, there's RAIntegration calls in this file.

src/frontend-common/fullscreen_ui.cpp
```diff
@@ -42,10 +43,6 @@
 #include <unordered_map>
 Log_SetChannel(FullscreenUI);
 
-#ifdef WITH_CHEEVOS
-#include "achievements.h"
```
stenzek, line 46: Ditto

PR #2925, opened 2022-11-21
Spanish (Spain) translation update

IlDucci:
Updating the Spanish (Spain) translation with the latest changes.

Please remember that the shared "qtbase_es.qm" is still not being duplicated as "qtbase_es-es.qm" within the release files, and thus, the generic translations (for buttons like Close/OK/Cancel) are still not being applied. Thank you.

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_es-es.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_es-es.ts
@@ -112,121 +112,187 @@
         <translation>Configuración global</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


stenzek:
> Please remember that the shared "qtbase_es.qm" is still not being duplicated as "qtbase_es-es.qm" within the release files, and thus, the generic translations (for buttons like Close/OK/Cancel) are still not being applied. Thank you.

Should be fixed in https://github.com/stenzek/duckstation/commit/b0e9208950a8d83072bf968eebc7c1dd88d2b971

PR #2941, opened 2023-01-16
"Set Dumped VRAM Write Alpha Channel" should be true when resetting advanced settings

bslenul:
Hey, currently when you click "Reset To Default" in the advanced settings, "Set Dumped VRAM Write Alpha Channel" will be disabled, I'm guessing this is a mistake since it's set to true everywhere else in the code by default.

There's also something wrong with the "PGXP Depth Clear Threshold" setting, if you start on a clean install (or if you delete "settings.ini"), it will be set to "4096" in the UI and "1228800" inside "settings.ini", but if you click "Reset To Default" it will be set to "300", which seems to be the desired value: https://github.com/stenzek/duckstation/blob/591e8b5b7a9e1c503df5674b0b0d0203d7eef6a7/src/core/settings.h#L430

I'm not sure to understand why that setting is multiplied/divised by 4096: https://github.com/stenzek/duckstation/blob/591e8b5b7a9e1c503df5674b0b0d0203d7eef6a7/src/core/settings.h#L284-L291 so I don't want to mess with it, too afraid to break something :p I just wanted to warn you about this!


stenzek:
Good catch, fixed in https://github.com/stenzek/duckstation/commit/55f716e69eeb7f0e2a6561913bb0c7759e6df4c9.

PR #2946, opened 2023-01-30
Add "Advanced Settings" option for rounding texture coordinates from an upscaled internal resolution

Superstarxalien:
### The problem

In order to prevent issues with displaying textures in games with upscaled internal resolutions, DuckStation floors all texture coordinates whenever vertex offset values are multiplied according to the internal resolution upscaling option. To my understanding, at least.

Flooring the texture coordinates doesn't work best for certain games, however. The Crash Bandicoot series of platformers shows jagged edges on the textures of crates whenever the game is upscaled, which is a pretty noticeable error throughout the entire games. For these games, rounding the texture coordinates instead of flooring them provides more "accurate" behavior, in regards to looking the closest to what the game looks like with its internal resolution unchanged.

These are screenshots I took of the games _Crash Bandicoot_, _Crash Bandicoot 2: Cortex Strikes Back_, and _Crash Bandicoot: WARPED_. Beyond the settings mentioned in each screenshot, I also enabled the "True Color" option to disable dithering.

</img></img></img>
_1x Internal Resolution_

</img></img></img>
_4x Internal Resolution (texture coordinates are floored, normal DuckStation behavior)_

</img></img></img>
_4x Internal Resolution (texture coordinates are rounded, modified DuckStation behavior)_

### What this commit does

This commit adds an option to force texture coordinates to always be rounded even when the internal resolution of the game is upscaled, found in the "Advanced Settings" category. While it would seem like a no-brainer to force this setting to be enabled for the games that look better with it, such as _Crash Bandicoot_...

</img>

It isn't actually perfect. Errors like these can be fixed simply by enabling PGXP, but, of course, that only benefits the people that wish to play with PGXP on. For such users, however, this option would be a no-brainer to enable.

P.S. the above screenshots were all with PGXP off, the game only looks bad right about there

Thanks to @ManDude for tipping me off on this problem and its (band-aid) solution.

### DIFF
--- a/src/core/gpu_hw.cpp
+++ b/src/core/gpu_hw.cpp
@@ -73,6 +73,7 @@ bool GPU_HW::Initialize()
   m_chroma_smoothing = g_settings.gpu_24bit_chroma_smoothing;
   m_downsample_mode = GetDownsampleMode(m_resolution_scale);
   m_disable_color_perspective = m_supports_disable_color_perspective && ShouldDisableColorPerspective();
+  m_round_upscale_coordinates = g_settings.gpu_round_upscale_coordinates;
 
   if (m_multisamples != g_settings.gpu_multisamples)
   {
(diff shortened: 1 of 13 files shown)


stenzek:
Yeah, that's why I never did myself.. my concern would be that it breaks more than it fixes.

I suppose you could try limiting it to 3D polygons (either by ignoring it for sprites, or using the Z from PGXP).

Superstarxalien:
so after some investigation, the glitch that occurs in crash 1's title screen when enabling this option can probably be fixed

that title screen is made up by `Rect16x16TexBlend` gpu calls, or in other words rectangle primitives; they're about the only graphics that get really messed up with this option on, so I figure that if the emulator checks for those and sets the texture coordinates to floor on those only it wouldn't be an issue anymore, looking for your thoughts though

stenzek:
Right, that's why I said "ignoring it for sprites" ;)

Superstarxalien:
well this is certainly how you can tell I'm less experienced in this subject matter lol
yeah I would go for that but I don't really know how to do it, so I wanted to ask you on that

stenzek:
I ended up implementing this independently in https://github.com/stenzek/duckstation/commit/8455c9c3c6f099425765aec4c73789c13a557065.

Added all the plumbing/tracking for separate pipelines for sprites, and... it seems I don't get the glitches shown above? Which is weird. Hopefully it's not vendor-specific, as `round()` has caused problems there before... but hopefully not `roundEven()`.

Anyway, that code can be re-used for separate texture filters for sprites.

PR #2952, opened 2023-02-25
GDBServer: Fix emulator deadlock on client connection

boricj:
The GDB server has a regression bisected down to 8d7aea5e19859ed483699cc4a5dbd47165c7be8b: DuckStation hangs when a client connects and the GDB connection tries to pause a running emulation, or at least it hangs every single time on my system.

I'm no Qt expert and I don't know what's going on (I also have hangs when I try to open the CPU Debugger window if the emulator isn't paused, so something's definitely fishy), but migrating the GDB stuff to the CPU emulation thread no longer causes DuckStation to hang while playing around with the GDB connection. We can't mess around with the emulated CPU if it's running anyway, so I don't expect this to be a problem.

While I'm here, let's also add an option toggle for enabling the GDB server in the Debug menu for visibility. It only modifies the setting and does not start/stop the GDB server on-the-fly, so it takes effect on the next power-on of the console.

### DIFF
--- a/src/duckstation-qt/gdbconnection.cpp
+++ b/src/duckstation-qt/gdbconnection.cpp
@@ -7,28 +7,31 @@
 #include "qthost.h"
 Log_SetChannel(GDBConnection);
 
-GDBConnection::GDBConnection(QObject* parent, int descriptor) : QThread(parent), m_descriptor(descriptor)
+GDBConnection::GDBConnection(GDBServer* parent, intptr_t descriptor) : QTcpSocket(parent), m_descriptor(descriptor)
 {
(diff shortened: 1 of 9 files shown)


boricj:
I _think_ I've finally managed to make Qt shut up about `moveToThread()` and reparenting children across threads stuff (I'm really a Qt noob). Hopefully I've sorted out this PR without introducing more issues in the process.

stenzek:
Yeah, cross-thread stuff in Qt definitely gets tricky.

Refactoring the server class to use BSD sockets instead was on my never-ending todo list, but obviously haven't got around to doing so.

PR #2962, opened 2023-04-26
UI QT: set helpText height to 136

charlesthobe:
now the tooltip for forcing 24-bit color depth fits exactly. 
Note: each new line in the standard font/formatting takes exactly additional 18px to fit.


stenzek:
Does this still fit on 768p/720p displays? If it doesn't already, then whatever, but some people are still using such displays in 2023 :/

CharlesThobe:
It doesn't already, but my pr makes it worse for those displays.

charlesthobe:
Closing so it doesn't clutter PRs page.

PR #2968, opened 2023-05-02
Default to X11 when supported

CharlesThobe:
This should be reverted once QT on Wayland gets fixed.

### DIFF
--- a/src/duckstation-qt/CMakeLists.txt
+++ b/src/duckstation-qt/CMakeLists.txt
@@ -252,3 +252,7 @@ if(APPLE)
   endif()
 endif()
 
+# Just for defaulting to X11, this and related code should be removed when QT fixes their Wayland implementation.
+if (USE_X11)
+  target_compile_definitions(duckstation-qt PRIVATE USE_X11)
+endif()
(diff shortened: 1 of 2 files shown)


CharlesThobe:
Also I can't get `Log_InfoPrintf()` to work.

stenzek:
What's the issue with Qt and wayland? (apart from the latter being fundamentally stupid and a mess, where you can't do simple things like set a window position)

CharlesThobe:
[Screencast from 2023-05-02 08-28-12.webm](https://user-images.githubusercontent.com/95266557/235586281-3f1f66d8-db63-4088-a32c-a0976ae23d76.webm)
This is the problem, on two very different laptops as well one with intel HD Graphics 3000 and another with intel HD Graphics 3000, same with nouveau drivers on both laptops (GT-520m and GT-740m respectively)
Also different distros Fedora and Debian and an Arch chroot on debian.

CharlesThobe:
I can test on an additional laptop with an intel Iris Xe Graphics, because I find it peculiar that nobody reported this issue.

CharlesThobe:
Also happens on pcsx2 and Dolphin file browser (under certain circumstances) when using wayland.

poorly written bug report: [KDE bug: 468172](https://bugs.kde.org/show_bug.cgi?id=468172)

CharlesThobe:
The trigger is starting the VM while on wayland, nothing else.

Affects both grid and list views.

CharlesThobe:
The rest of the UI is usable.

one way to avoid the bug while on wayland is enabling the "Render To Separate Window" option.

PR #2970, opened 2023-05-02
CMake: display warning when building without Wayland on systems that support Wayland

CharlesThobe:


CMakeLists.txt
```diff
@@ -55,6 +55,11 @@ if((LINUX OR FREEBSD) AND NOT ANDROID)
   option(USE_EVDEV "Support EVDev controller interface" OFF)
 endif()
 
+# Error out when building without Wayland on a system that has QT built with Wayland support.
+if (SUPPORTS_WAYLAND AND NOT USE_WAYLAND AND IS_DIRECTORY "${Qt6_DIR}/../Qt6WaylandClient")
+message(FATAL_ERROR "It seems Qt was compiled with Wayland support, you should set USE_WAYLAND")
```
stenzek, line 60: message should be indented

CharlesThobe:
Wait a sec before merging

CharlesThobe:
It configures fine with -DUSE_WAYLAND=OFF here

CharlesThobe:
> It configures fine with -DUSE_WAYLAND=OFF here

Thing is, even if you don't have developer package `qt6-wayland-dev` installed i.e no `$base/cmake/Qt6WaylandClient/` exists, but have the user package `qt6-wayland` it still built fine, but qt still supported wayland even then, apparently you can't toggle a specific QPA_PLATFORM support. i.e QT and Wayland support are "inseparable", if you have the QPA plugin lib it will use it regardless of anything else.

CharlesThobe:
Which makes me wonder. what is the `qt6-wayland-dev` package is for? 😄 

also almost the same story with `libqt6svg6` and `libqt6svg6-dev`, duckstation doesn't need the latter for building but needs the former anyway on runtime to display svg icons, I guess they're all optional "QT modules/plugins".

PR #2974, opened 2023-05-10
Upstream Netplay Changes

HeatXD:
- wire up netplay chat and also created a channel for non-control type messages
- removed netplay ping widget from the main window

### DIFF
--- a/src/core/netplay.cpp
+++ b/src/core/netplay.cpp
@@ -523,16 +523,16 @@ void Netplay::HandleEnetEvent(const ENetEvent* event)
       {
         HandleControlMessage(player_id, event->packet);
       }
-      else if (event->channelID == ENET_CHANNEL_SESSION)
-      {
-        HandleSessionMessage(player_id, event->packet);
-      }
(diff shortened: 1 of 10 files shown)


src/core/netplay.cpp
```diff
@@ -48,13 +48,25 @@ enum class ControlMessage : u32
   SynchronizeComplete,
 };
 
+enum class SessionMessage : u32
+{
+  ChatMessage,
```
stenzek, line 53: Any reason why this couldn't be a control message? (would save allocating another channel)

HeatXD: Was simply an assumption. If it being a control message wont hinder future plans than it isnt a problem. it can indeed just be a control message then. will fix when i wakeup tomorrow 👍

stenzek: Perhaps my naming wasn't the best, I basically split into "ggpo" and "non-ggpo" (control) packet types. The latter doesn't require the ggpo session to be valid, so it can be used while resynchronizing.

PR #2979, opened 2023-05-17
Netplay: Session Password

HeatXD:


src/core/netplay.cpp
```diff
@@ -1861,12 +1871,13 @@ void Netplay::SetInputs(Netplay::Input inputs[2])
 
 bool Netplay::CreateSession(std::string nickname, s32 port, s32 max_players, std::string password)
 {
-  // TODO: Password
+  s_local_session_password = password;
```
stenzek, line 1874: std::move(password)

src/core/netplay.cpp
```diff
@@ -1885,8 +1896,8 @@ bool Netplay::CreateSession(std::string nickname, s32 port, s32 max_players, std
 
 bool Netplay::JoinSession(std::string nickname, const std::string& hostname, s32 port, std::string password)
 {
-  // TODO: Password
-
+  s_local_session_password = password;
```
stenzek, line 1899: same here

PR #2986, opened 2023-07-08
Linux: fix potentially unsafe screensaver inhibitor

charlesthobe:
The dbus behavior according to their documentation is as follows:
1- Calling `dbus_bus_get()` gets you a public connection, there can only be one of those.
2- This connection should remain open for the rest of the life of the program and is unclosable.
3- Calling `dbus_connection_unref(ReferenceToThePublicConnection)` can result in UB.
4- If the connection is severed in case of dbus daemon crash for example, the dbus lib automatically closes the program (duckstation closes) unless `dbus_connection_set_exit_on_disconnect(ReferenceToThePublicConnection, false)` is called.

All of that means that the connection will always be alive and therefor the inhibitor cookie will always be valid until used, meaning that the cookie shouldn't be discarded until it's used successfully in removing the screensaver inhibitor (reasons for failure of doing so are things like the CPU being under a too heavy load that it takes more time than `DBUS_TIMEOUT_USE_DEFAULT` for duckstation to receive a requested message from dbus), so even then the cookie should be kept for reuse and duckstation should be prevented from requesting another inhibitor, until the current inhibitor is gone.

The PR addresses all of that.

### DIFF
--- a/src/frontend-common/platform_misc_unix.cpp
+++ b/src/frontend-common/platform_misc_unix.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: (GPL-3.0 OR CC-BY-NC-ND-4.0)
 
 #include "common/log.h"
+#include "common/scoped_guard.h"
 #include "common/string.h"
 #include "input_manager.h"
 #include "platform_misc.h"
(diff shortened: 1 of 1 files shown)


charlesthobe:
I opted not to use `dbus_connection_set_exit_on_disconnect()` btw

charlesthobe:
Even if the cookie becomes invalid due to success in inhibition, but too late of a response, having the function consistently fail is better than having the potential of inflating the daemon with inhibitors, imo.

charlesthobe:
I have an improvement on my mind, a couple of hours until I return home.

charlesthobe:
> I have an improvement on my mind, a couple of hours until I return home.

Partially addressed.

charlesthobe:
I think this is mergeable currently, doing more would overcomplicate it for a "problem" that's just a red herring.

PR #2988, opened 2023-07-11
Fix compile error in evdev_input_source.cpp

rtissera:
Trivial fix.


stenzek:
No, because it was pointless - didn't support automatic mapping, required messing around with device node permissions to use, etc.

NoGUI is not intended for users, it's missing a bunch of features, including the updater, error/confirm messages, relative mouse mode, etc.. That's why I don't ship it in any releases - it's there for bringing up new platforms, e.g. I used it while I was writing the RISC-V dynarec, because Qt6 isn't available there yet.

Qt can launch straight into big picture mode for some time now.

PR #2994, opened 2023-08-15
update Simple-Chinese translation to latest.

zkdpower:
update Simple-Chinese translation to latest.
besides, 
1. please add string of "No such file or directory" in OSD message for translating. 
2. Please consider replacing icon of "Trace" which under the "CPU Debugger" . it has no transparent effect like the other icons.


stenzek:
> please add string of "No such file or directory" in OSD message for translating.

FYI, this comes from the operating system, not the application (FormatMessage).

PR #2995, opened 2023-08-19
Update audiosettingswidget.cpp

xujibbs:


stenzek:
Cubeb is the name of the backend - that's why it's not translatable.

xujibbs:
Will you make "50 ms" translatable?

stenzek:
> Will you make "50 ms" translatable?

Added in https://github.com/stenzek/duckstation/commit/bfa792ddbff11c102521124f235ccb310cac6e6a

PR #2997, opened 2023-08-20
Fix `HostReg` enum cast outside valid range

j8r:
LLVM 16 errors by reporting that the valid range of `HostReg` is between 0 and 15.

I don't have enough knowledge if this is the best fix. ~Using a different type accepting the `16` value may be another option~. I changed to an unsigned type.

Here is the full error:
```
/duckstation/src/core/cpu_recompiler_types.h:76:37: error: integer value 16 is outside the valid range of values [0, 15] for this enumeration type [-Wenum-constexpr-conversion]
constexpr HostReg HostReg_Invalid = static_cast<HostReg>(HostReg_Count);
                                    ^
```

### DIFF
--- a/src/core/cpu_recompiler_types.h
+++ b/src/core/cpu_recompiler_types.h
@@ -71,7 +71,7 @@ using CodeEmitter = Xbyak::CodeGenerator;
 using LabelType = Xbyak::Label;
 enum : u32
 {
-  HostReg_Count = 16
+  HostReg_Count = 15
 };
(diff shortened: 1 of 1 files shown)


stenzek:
I had already changed this as part of https://github.com/stenzek/duckstation/commit/bf15d13eb7398ed635e9b54960ac110c709ce0f6 yesterday for clang-cl compatibility, just hadn't pushed it yet.

PR #3006, opened 2023-09-03
Update Russian translation

shikulja:


stenzek:
Not sure why you need to ignore .qm? The only qm files should be in directories which are already ignored.

shikulja:
Before committing, I create a qm file for testing. He often asks to be added to commits too
image

stenzek:
Sounds like a local issue then. Please drop that commit.

shikulja:
GitHub Desktop + win11 nothing special

PR #3009, opened 2023-09-05

andercard0:
Alterado/Atualizado
- Opções do modo big duck mode. adicionadas

Missing Strings or not translateable:
- Rewind is not enable. Please note that enabling rewind may significantly increase system requriments. (in configuration settings)
- Rendering (in video options)

Text title
- PGXP (Precision Geometry Transform Pipeline)
- Input sources (in controller options tab)
- Not inverted

Shortcut configuration options
- Settings and operations (Text title)
- Settings (Text title)
- Account
- Current game

@stenzek let me know how to do it when you can so I do it myself next time a see a untranslatable string please?!


stenzek:
Should be fixed in https://github.com/stenzek/duckstation/commit/99c30b1e80e5a35077077be9f0fa665bd7d9fe4c

PR #3072, opened 2023-12-18

zkdpower:
i found a rewind bug, i.e——pressing the rewind hotkey unreleasly to Original place will cause the app crash/dead.


zkdpower:
in latest 0.6234 version, pressing the rewind hotkey unreleasly now result a new break, and tips
Assertion failed:'block>next_block_in_page' in function CreateBlock (cpu_code_cache.cpp:430)

stenzek:
Should be fixed now.

PR #3078, opened 2023-12-26
Update ts files

andercard0:
Remove obsolete strings for all available language files.


stenzek:
Might need to reset your fork, seems to have made a dog's breakfast of some of my earlier commits (probably from the force pushes..).

Committed in https://github.com/stenzek/duckstation/commit/489235f48a593f208e722c4c0ce67a31736d8bec

andercard0:
> Might need to reset your fork, seems to have made a dog's breakfast of some of my earlier commits (probably from the force pushes..).
> 
> Committed in https://github.com/stenzek/duckstation/commit/489235f48a593f208e722c4c0ce67a31736d8bec

Strange, this one is already the new one. I will do it again later.

PR #3089, opened 2024-01-19
Add debanding option to improve gradient smoothness with truecolor

Calinou:
This option only has an effect if true color is also enabled.

Most PS1 games don't exhibit that much banding with true color, but there are particular cases where flickering induced by a lack of color precision is visible, such as Driver in its nighttime levels. You may also notice more banding if you play with bilinear texture filtering enabled, in which case debanding will help.

Compared to the existing deband post-processing shader, this works with internal rendering which makes it both faster and more precise. No loss of detail occurs, and no tweaking of thresholds is required.

The option is disabled by default as its effect is subtle in most games, and it increases the file size of screenshots due to the subtle dithering pattern.

The debanding algorithm used is very fast (only a handful of instructions), and is also the one that happens to be used in Godot :slightly_smiling_face: 

## Preview

*Click to view at full size.*

Before | After
-|-
driver_without_debanding | driver_debanding

This kind of flickering on the parking lot's floor and ceiling completely goes away once debanding is enabled:

https://github.com/stenzek/duckstation/assets/180032/ee325ce0-ddbc-41a3-8452-7472e8de7ff9

### DIFF
--- a/src/core/fullscreen_ui.cpp
+++ b/src/core/fullscreen_ui.cpp
@@ -3997,6 +3997,10 @@ void FullscreenUI::DrawDisplaySettingsPage()
                     FSUI_CSTR("Disables dithering and uses the full 8 bits per channel of color information."), "GPU",
                     "TrueColor", true, is_hardware);
 
+  DrawToggleSetting(bsi, FSUI_CSTR("True Color Debanding"),
+                    FSUI_CSTR("Applies modern dithering techniques to further smooth out gradients when true color is enabled."), "GPU",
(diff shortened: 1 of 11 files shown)


stenzek:
Sounds good, thanks for this!

Some minor UI changes, like having true colour disabled should disable banding, but I'll sort that out post-merge.

I also wonder if it would be better to apply after the texture modulation step, rather than to the two independently, since that would increase the visibility of the dithering pattern.

Linux CI fail is unrelated to these changes.

PR #3092, opened 2024-01-21
VulkanDevice: Accelerate texture downloads with `VK_EXT_external_memory_host`

Wunkolo:
With [VK_EXT_external_memory_host](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_external_memory_host.html), host memory can be directly imported into a `VkDeviceMemory`/`VkBuffer` without the need to allocate or copy from staging memory. When [minImportedHostPointerAlignment](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceExternalMemoryHostPropertiesEXT.html#_members) matches the page-size of the system, then basically _any_ host pointer can be imported into Vulkan by importing the entire span of `PAGE_SIZE`-chunks of memory into Vulkan and offsetting into it.

In the case of large texture downloads like upscaled screenshots, this saves a good chunk of memory from having to be allocated by directly writing texture-data into `void* out_data`.

This could work for any kind of CPU<->GPU transfer too.

### DIFF
--- a/src/util/vulkan_device.cpp
+++ b/src/util/vulkan_device.cpp
@@ -2952,6 +2952,90 @@ void VulkanDevice::RenderBlankFrame()
   InvalidateCachedState();
 }
 
+bool VulkanDevice::TryImportHostMemory(const void* data, u32 data_size, VkBufferUsageFlags buffer_usage,
+                                       VkDeviceMemory* out_memory, VkBuffer* out_buffer, u32* out_offset)
+{
+  if (!m_optional_extensions.vk_ext_external_memory_host)
+    return false;
+
+  // Align to the nearest page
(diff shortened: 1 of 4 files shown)


stenzek:
I suspect creating a new buffer for every download may negate the benefit of going direct to client memory in the first place.

The ideal way of handling this would probably to have a "DownloadTexture" abstraction like I did for PCSX2, which either creates a shadow buffer, or imports the client VRAM buffer.

stenzek:
I've implemented aforementioned download texture concept now, so we can close this PR. I did cherry-pick the enablement commits, however, so thanks for that!

PR #3113, opened 2024-02-05
Qt: Icon updates

kamfretoz:
This PR backport the updated icons from PCSX2

Here's some preview:
- New cheat icons
image

- New Analog OSD icon
Screenshot_20240205_201403

- New Advanced icon 
image

- New memcard and gamepad icon for BPM
Screenshot_20240205_203932

And some refinements here and there.


stenzek:
Could you add whatever you changed in font awesome to promptfont instead? Not a fan of modifying the binary file without any way of rebuilding it - the changes will get lost.

Could use the same code point, just swap the macro out for PF instead of FA.

kamfretoz:
Updated, FontAwesome remains untouched.

Screenshot_20240207_211642

zkdpower:
this icon's white background is different.

kamfretoz:
I'll look into it

kamfretoz:
All good now.

Screenshot_20240215_214556

src/duckstation-qt/debuggerwindow.ui
```diff
@@ -366,98 +361,91 @@
...
    <property name="icon">
-    <iconset resource="resources/resources.qrc">
-     <normaloff>:/icons/debug-step-out.png</normaloff>:/icons/debug-step-out.png</iconset>
+    <iconset theme="debug-step-out-line"/>
    </property>
    <property name="text">
     <string>Step Out</string>
    </property>
    <property name="toolTip">
-    <string>Step O&amp;ut</string>
```
stenzek, line 376: These should be on the text instead of the tooltip, I'm guessing.

src/duckstation-qt/debuggerwindow.cpp
```diff
@@ -251,10 +251,10 @@ void DebuggerWindow::onCodeViewContextMenuRequested(const QPoint& pt)
   menu.addAction(QStringLiteral("0x%1").arg(static_cast<uint>(address), 8, 16, QChar('0')))->setEnabled(false);
   menu.addSeparator();
 
-  QAction* action = menu.addAction(QIcon(":/icons/media-record@2x.png"), tr("Toggle &Breakpoint"));
+  QAction* action = menu.addAction(tr("Toggle &Breakpoint"));
   connect(action, &QAction::triggered, this, [this, address]() { toggleBreakpoint(address); });
 
-  action = menu.addAction(QIcon(":/icons/debug-run-cursor.png"), tr("&Run To Cursor"));
+  action = menu.addAction(tr("&Run To Cursor"));
```
stenzek, line 257: Was removing the icons here intentional? You can use QIcon::fromTheme to create a light/dark icon.

kamfretoz: Whoops, i thought we don't need it since the theme is already defined on the .ui file

stenzek: This one's being created in the CPP file, not ui

kamfretoz: Just noticed that its for the context menu icon, fixed:

image

src/duckstation-qt/debuggerwindow.ui
kamfretoz: That does it.

image

PR #3136, opened 2024-03-17
Update fullscreen_ui.cpp

xujibbs:
make more strings translatable


stenzek:
Just defining the strings is insufficient, you need to change the sites where they're actually used (using the appropriate `FSUI_` macro). Then you can run the `generate_fullscreen_ui_translation_strings.py` script (under `scripts`) to update the string table.

xujibbs:
I was unable to change the sites, if possible, please help to fulfil the missing strings. Duckstation is a great Artwork, hoping to own a perfect localization.

stenzek:
Done in https://github.com/stenzek/duckstation/commit/3085172ffd20b8c3e4f87a5758448302f4aa209a.

PR #3150, opened 2024-04-04
[Cheat DB] Update

andercard0:
Various codes from discord cheat-workshop by: Teh69thSpartan added. cheat Credits :hugopocked


andercard0:
As discussed in discord, these codes need a bit of refactoring which will be done properly by Pugsy in the future, closing it since it's not much useful right now.

PR #3171, opened 2024-04-12
Renamed Mem Scanner Op Text&added Virtual Address

PugsyMAME:
Renamed the Memory Scanner Operator Text for clarity. Replaced '...' with 'Value' and 'Previous' with 'Previous Result'.
Also changed "Any Value" to "Any Value ('New Search' or 'Reset Result Value')", "Any Value" should really be at the top of the index and the Previous Results should come before the other Value operators. But that's a job for another time.

The Virtual Address extra will display the virtual address of DuckStation's PSX RAM so it can be easily edited with the likes of HXD or another tool where the location of the virtual memory needs to be known.

### DIFF
--- a/src/core/cheats.h
+++ b/src/core/cheats.h
@@ -204,22 +204,22 @@ class MemoryScan
 public:
   enum class Operator
   {
-    Equal,
-    NotEqual,
-    GreaterThan,
-    GreaterEqual,
-    LessThan,
-    LessEqual,
-    IncreasedBy,
-    DecreasedBy,
-    ChangedBy,
-    EqualLast,
-    NotEqualLast,
-    GreaterThanLast,
-    GreaterEqualLast,
+    Any,
     LessThanLast,
     LessEqualLast,
-    Any
+    GreaterThanLast,
+    GreaterEqualLast,
+    NotEqualLast,
+    EqualLast,
+    DecreasedBy,
+    IncreasedBy,
+    ChangedBy,
+    Equal,
+    NotEqual,
+    LessThan,
+    LessEqual,
+    GreaterThan,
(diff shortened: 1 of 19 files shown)


PugsyMAME:
Ideally I'd like to change the order as Previous Results comparisons are more useful and should be before Value comparisons. 

Exception: Any Value - is ideal to be used as for 'New Search' and should be at the top.  It's also useful when you are not sure if you've messed up and you don't want to screw up the current search.

Can't seem to locate how the index of the "Operator:" is tested. For the above PR I did the change semi-manually....I don't know if there's some automated method of updating the location line in the 16 translation files - if there's not it could be quite error prone I guess.

Existing Order (but new names):-

```
Equal to Value
Not Equal to Value
Greater Than Value
Greater or Equal to Value
Less Than Value
Less or Equal to Value
Increased By Value
Decreased By Value
Changed By Value
Equal to Previous Result (Unchanged Value)
Not Equal to Previous Result (Changed Value)
Greater Than Previous Result
Greater or Equal to Previous Result
Less Than Previous Result
Less or Equal to Previous Result
Any Value ('New Search' or 'Reset Result Value')
```

Better Order:-

```
Any Value ('New Search' or 'Reset Result Value') 
Greater Than Previous Result
Greater or Equal to Previous Result
Less Than Previous Result
Less or Equal to Previous Result
Not Equal to Previous Result (Changed Value)
Equal to Previous Result (Unchanged Value)
Increased By Value
Decreased By Value
Changed By Value
Equal to Value
Not Equal to Value
Greater Than Value
Greater or Equal to Value
Less Than Value
Less or Equal to Value
```

PR #3172, opened 2024-04-12
Renamed Mem Scanner Op Text&added Virtual Address

PugsyMAME:
Renamed the Memory Scanner Operator Text for clarity. Replaced '...' with 'Value' and 'Previous' with 'Previous Result'.
Also changed "Any Value" to "Any Value ('New Search' or 'Reset Result Value')"

Reordered the Memory Scanner Operator Text into a more intuitive order:

Any Value ('New Search' or 'Reset Result Value') 
Less Than Previous Result
Less or Equal to Previous Result
Greater Than Previous Result
Greater or Equal to Previous Result
Not Equal to Previous Result (Changed Value)
Equal to Previous Result (Unchanged Value)
Increased By Value
Decreased By Value
Changed By Value
Equal to Value
Not Equal to Value
Greater Than Value
Greater or Equal to Value
Less Than Value
Less or Equal to Value

The Virtual Address extra will display the virtual address of DuckStation's PSX RAM so it can be easily edited with the likes of HXD or another tool where the location of the virtual memory needs to be known.

### DIFF
--- a/src/duckstation-qt/memoryscannerwindow.cpp
+++ b/src/duckstation-qt/memoryscannerwindow.cpp
@@ -91,9 +91,7 @@ MemoryScannerWindow::MemoryScannerWindow() : QWidget()
   m_ui.setupUi(this);
   connectUi();
 
-  char tempvar[256];
-  sprintf(tempvar, "Address of RAM for HxD Usage: 0x%p", Bus::g_ram);
-  m_ui.cheatEngineAddress->setText(tempvar);
+  m_ui.cheatEngineAddress->setText(tr("Address of RAM for HxD Usage: 0x%1").arg(reinterpret_cast<qulonglong>(Bus::g_unprotected_ram), 16, 16, QChar('0')));
 }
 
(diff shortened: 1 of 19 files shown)


src/duckstation-qt/memoryscannerwindow.cpp
```diff
@@ -90,6 +90,10 @@ MemoryScannerWindow::MemoryScannerWindow() : QWidget()
 {
   m_ui.setupUi(this);
   connectUi();
+
+  char tempvar[256];
+  sprintf(tempvar, "Address of RAM for HxD Usage: 0x%p", Bus::g_ram);
+  m_ui.cheatEngineAddress->setText(tempvar);
```
stenzek, line 96: Please don't use sprintf, it's unsafe (even if it's not going to overflow).

```suggestion
  m_ui.cheatEngineAddress->setText(tr("Address of RAM for HxD Usage: 0x%1").arg(reinterpret_cast<qulonglong>(Bus::g_unprotected_ram), 16, 16, QChar('0')));
```

PR #3177, opened 2024-04-15
Negcon rumble

sonik-br:
This PR adds a new device: Negcon with rumble.
There's no QT interface for it, but the "Automatic Mapping" should work.
Rumble motors should also be automatic mapped to the current standard left and right motors.

It's also possible to toogle between a digital pad and the new Negcon mode by pressing the "analog button".
I've added this as at least one game refused to pass the initial screen when in NegconRumble mode. After the screen, the mode change worked as expected.

@CookiePLMonster helped with testing a lot of games.
Here's the list of some of the games that supports this device.
* Advan Racing
* CMR1
* CMR2.0
* Gran Turismo 1
* NASCAR '99 Legacy
* NASCAR 2000
* NASCAR 2001
* Rally de Africa
* Rally de Europe
* Side by Side Special 2000
* TOCA1
* TOCA2
* TOCA WTC
* NFS4

### DIFF
--- a/src/core/negcon_rumble.cpp
+++ b/src/core/negcon_rumble.cpp
@@ -1,0 +1,788 @@
+// SPDX-FileCopyrightText: 2019-2022 Connor McLaughlin <stenzek@gmail.com> and contributors.
+// SPDX-License-Identifier: (GPL-3.0 OR CC-BY-NC-ND-4.0)
+
+#include "negcon_rumble.h"
+#include "IconsFontAwesome5.h"
+#include "common/assert.h"
+#include "common/log.h"
+#include "host.h"
+#include "system.h"
+#include "util/state_wrapper.h"
+#include <array>
+#include <cmath>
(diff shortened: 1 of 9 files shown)


src/core/negcon_rumble.cpp
```diff
@@ -0,0 +1,813 @@
...
+  //  if (sub_index >= m_axis_state.size())
+  //    return 0.0f;
+
+  //  return static_cast<float>(m_axis_state[sub_index]) * (1.0f / 255.0f);
+  //}
+  //else
+  //{
+  //  const u32 bit = s_button_indices[index];
+  //  return static_cast<float>(((m_button_state >> bit) & 1u) ^ 1u);
+  //}
```
stenzek, line 172: Please don't leave code like this commented out. If it's not needed, remove it. There were a few others below and in the header too.

sonik-br: Done.

PR #3195, opened 2024-05-04
Update fullscreen_ui.cpp

xujibbs:
make strings translatable


stenzek:
Pretty sure I said this last time, but simply adding the strings has no effect. You need to add `FSUI_CSTR` at the use site.

xujibbs:
Yes, I remember your words. let me do again.

xujibbs:
Hoping the similar strings in pcsx2/ImGui/FullscreenUI.cpp would be also updated soon to be translatable.

PR #3208, opened 2024-05-17
Bump GitHub Actions and Ubuntu

Addin:
- Update actions/checkout from version 4.1.4 to 4.1.6
- Update actions/upload-artifact from version 4 to 4.3.3
- Update actions/download-artifact from version 4 to 4.1.7

### DIFF
--- a/.github/workflows/rolling-release.yml
+++ b/.github/workflows/rolling-release.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: windows-2022
     timeout-minutes: 120
     steps:
-    - uses: actions/checkout@v4.1.4
+    - uses: actions/checkout@v4.1.6
       with:
         fetch-depth: 0
 
@@ -98,7 +98,7 @@ jobs:
         "C:\Program Files\7-Zip\7z.exe" a -r duckstation-windows-x64-release.zip ./bin/x64/*
 
     - name: Upload x64 release artifact
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v4.3.3
       with:
         name: "windows"
         path: "duckstation-windows-x64-release*.zip"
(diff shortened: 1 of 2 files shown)


.github/workflows/rolling-release.yml
```diff
@@ -385,19 +385,19 @@ jobs:
...
 
   create-release:
     needs: [windows-build, windows-arm64-build, linux-build, linux-flatpak-build, macos-build]
-    runs-on: "ubuntu-22.04"
+    runs-on: ubuntu-24.04
     if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/dev'
     steps:
       - name: Download Artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/upload-artifact@v4.3.3
```
stenzek, line 400: This is wrong.

.github/workflows/rolling-release.yml
```diff
@@ -183,14 +183,14 @@ jobs:
...
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v4.3.3
       with:
         name: "windows-arm64"
         path: "duckstation-windows-arm64-release*.zip"
 
 
   linux-build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
stenzek, line 193: Linux builders have to stay on 22.04, because otherwise 24.04 becomes the minimum for the AppImage.

.github/workflows/rolling-release.yml
Addin: Indeed it is wrong, already fixed.

.github/workflows/rolling-release.yml
Addin: Got it, thanks for the heads up!
Reverted to ubuntu-22.04.

PR #3214, opened 2024-05-25
Fix README.md Linux build instructions

bemug:


stenzek:
Thanks for the PR, however the deps script was intended to work with relative paths, the fact that it didn't was a bug (fixed in https://github.com/stenzek/duckstation/commit/9eb18449a74c5b630ed98681dde3f8c9293fcdf7).

I don't really care either way too much about the example directory name. The expectation was that most users wouldn't compile a debug build, so the suffix was unwarranted, but if you want to add it, go nuts.

bemug:
Thank you for 9eb1844. I confirm it now works with relative paths.

In step 2 `scripts/linux/build-dependencies-linux.sh` must still be changed to `scripts/build-dependencies-linux.sh`. I guess the script have moved over time.

About the directory name in step 4, I changed it to use the same build directory as step 3 so that it works by copy-pasting. You're correct by saying step 4 could be something else than a release build.
It's up to you. Fell free to remove it if it's a bad idea.

I'll update this PR to reflect these changes.

PR #3215, opened 2024-05-26
Add memory contents to "Addr=" string in debugger

PugsyMAME:
Tweaks the format of the addr= string in the debugger/trace so that it includes the memory contents. It will also take the 2nd character of the opcode and if it's b or h it will size it accordingly. so
addr=800BC2D0
becomes
addr=800BC2D0[00000000]  or addr=800BC2D0[0000]  or addr=800BC2D0[00] 

This change amongst other things will facilitate a widescreen cheat scanner, that will take a lot of the work away from making widescreen cheats.

### DIFF
--- a/src/core/cpu_disasm.cpp
+++ b/src/core/cpu_disasm.cpp
@@ -434,8 +434,46 @@ void CPU::FormatComment(SmallStringBase* dest, const Instruction inst, u32 pc, c
     else if (std::strncmp(str, "offsetrs", 8) == 0)
     {
       const s32 offset = static_cast<s32>(inst.i.imm_sext32());
-      dest->append_format("{}addr={:08X}", dest->empty() ? "" : ", ",
-                          regs->r[static_cast<u8>(inst.i.rs.GetValue())] + offset);
+      u32 tempValueWord;
+      u16 tempValueHalfWord;
+      u8 tempValueByte;
+      bool returnValue=false;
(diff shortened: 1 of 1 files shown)


src/core/cpu_disasm.cpp
```diff
@@ -434,8 +434,46 @@ void CPU::FormatComment(SmallStringBase* dest, const Instruction inst, u32 pc, c
...
+      u32 tempValueWord;
+      u16 tempValueHalfWord;
+      u8 tempValueByte;
+      bool returnValue=false;
+      u32 tempAddress = (regs->r[static_cast<u8>(inst.i.rs.GetValue())] + offset);
+
+      SmallString instructionStr;
+      CPU::DisassembleInstruction(&instructionStr, pc, inst.bits);
+
+      if (instructionStr[1] == 'b')
```
stenzek, line 446: Rather than disassembling the instruction (slow), you can look at the instruction's `op`, and compare with `op == InstructionOp::lb || op == InstructionOp::sb`.

PugsyMAME: Thanks, that is indeed a lot easier. Didn't realise that InstructionOp:: existed. 

I will probably not include store opcode checks as the existing contents only matter during loads.

I'll clean it up (can save a few lines of code with that knowledge) and resubmit.

PugsyMAME:
Cleaned up and resubmitting

PR #3216, opened 2024-05-26
Add memory contents to "Addr=" string in debugger(Update]

PugsyMAME:
Tweaks the format of the addr= string in the debugger/trace so that it includes the memory contents. It will also take the 2nd character of the opcode and if it's b or h it will size it accordingly. so
addr=800BC2D0
becomes
addr=800BC2D0[00000000] or addr=800BC2D0[0000] or addr=800BC2D0[00]

This change amongst other things will facilitate a widescreen cheat scanner, that will take a lot of the work away from making widescreen cheats.

Changed code as per advice.


stenzek:
Merged in https://github.com/stenzek/duckstation/commit/f7025d6f33c2aed598e7e82fdecc02eeb774087d, just added a little further tidy up.

PR #3219, opened 2024-05-30
Bus: Expose shared memory on Unix systems

bemug:
On Windows, the shared memory handle is closed on Destroy() and therefore is accessible at runtime. On Unixes the shared memory was unlinked right after its creation, effectively making the shared memory unavailable from the outside world.
This commit brings feature parity to Unix systems. Allows the shared memory to be accessed at runtime from an external process for RAM inspection.

### DIFF
--- a/src/common/memmap.cpp
+++ b/src/common/memmap.cpp
@@ -63,8 +63,11 @@ void* MemMap::CreateSharedMemory(const char* name, size_t size, Error* error)
   return mapping;
 }
 
-void MemMap::DestroySharedMemory(void* ptr)
+void MemMap::DestroySharedMemory(void* ptr, const char* name)
 {
+  // Unused
+  (void)name;
+
   CloseHandle(static_cast<HANDLE>(ptr));
 }
 
@@ -319,8 +322,11 @@ void* MemMap::CreateSharedMemory(const char* name, size_t size, Error* error)
   return reinterpret_cast<void*>(static_cast<uintptr_t>(port));
 }
 
-void MemMap::DestroySharedMemory(void* ptr)
+void MemMap::DestroySharedMemory(void* ptr, const char* name)
 {
+  // Unused
+  (void)name;
+
   mach_port_deallocate(mach_task_self(), static_cast<mach_port_t>(reinterpret_cast<uintptr_t>(ptr)));
 }
 
@@ -487,9 +493,6 @@ void* MemMap::CreateSharedMemory(const char* name, size_t size, Error* error)
     return nullptr;
   }
 
-  // we're not going to be opening this mapping in other processes, so remove the file
-  shm_unlink(name);
-
   // use fallocate() to ensure we don't SIGBUS later on.
 #ifdef __linux__
   if (fallocate(fd, 0, 0, static_cast<off_t>(size)) < 0)
@@ -509,8 +512,9 @@ void* MemMap::CreateSharedMemory(const char* name, size_t size, Error* error)
   return reinterpret_cast<void*>(static_cast<intptr_t>(fd));
 }
 
-void MemMap::DestroySharedMemory(void* ptr)
+void MemMap::DestroySharedMemory(void* ptr, const char* name)
 {
+  shm_unlink(name);
   close(static_cast<int>(reinterpret_cast<intptr_t>(ptr)));
 }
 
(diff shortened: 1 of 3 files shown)


stenzek:
Sorry, I'm fundamentally against this.
1. If DS crashes, then it leaves a file hanging around.
2. It doesn't handle invalidating the recompiler code cache when you overwrite previously compiled blocks.

This is why I implemented PINE, which executes in the context of the CPU thread, and safely invalidates the rec as needed.

bemug:
I was not aware of PINE's existence. If you tell me it can achieve the same things without these limitations then that sounds pretty good. Thank you.

Do you have any example code available somewhere to use PINE ?

kamfretoz:
> Do you have any example code available somewhere to use PINE ?

Here is [PINE's repo](https://github.com/GovanifY/pine)

stenzek:
It's _fairly_ straightforward, you just send over a 4-byte packet length, 1-byte command type, and variable args over a unix domain socket in `$XDG_RUNTIME_DIR/duckstation.sock` (on Linux), or `127.0.0.1:28001` (on Windows).

So e.g. to write 0x12345678 to address 0x80001000, you would send:
```
uint32(13) // 0D 00 00 00 - command length
uint8(6) // 06 - command type, MsgWrite32
uint32(0x80001000) // 00 10 00 80 - memory address
uint32(0x12345678) // 78 56 34 12 - data
```

and get back:
```
uint32(0) // 00 00 00 00 - reply length
uint8(0) // 00 - success (fail is 0xFF)
```

You don't _need_ to handle the replies, can just ignore them, but you should read them out, otherwise your socket receive buffer will eventually fill, and the DS side will stop processing commands.

Full command list: 
https://github.com/stenzek/duckstation/blob/f1465ddf4323e5438452cb6ccaed581b8d2ee735/src/core/pine_server.cpp#L51-L70
https://github.com/stenzek/duckstation/blob/f1465ddf4323e5438452cb6ccaed581b8d2ee735/src/core/pine_server.cpp#L261-L526

This is the test script I used when writing the implementation: 
```python
import socket
import binascii
import platform
import os
   
def get_socket():
    if (platform.system() == "Windows"):
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        s.connect(("127.0.0.1", 28011))
    else:
        sdir = os.getenv("XDG_RUNTIME_DIR")
        if sdir is None:
            sdir = os.getenv("TEMPDIR", "/tmp")
        sfilename = f"{sdir}/duckstation.sock"
        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        s.connect(sfilename)
    return s
    

def print_hex_ascii(data):
    def to_printable_ascii(byte):
        return chr(byte) if 32 <= byte <= 126 else '.'

    for i in range(0, len(data), 8):
        chunk = data[i:i+8]
        hex_part = ' '.join(f'{byte:02x}' for byte in chunk)
        ascii_part = ''.join(to_printable_ascii(byte) for byte in chunk)
        print(f"{hex_part:<23} {ascii_part}")

        
def test_query(data):
    s = None    
    try:
        s = get_socket()
        for i in range(4):
            s.sendall(data)
            print(f"Sent data: {binascii.hexlify(data).decode('utf-8')}")
            response = s.recv(4096)
            print("Response:")
            print_hex_ascii(response)
    
    except Exception as e:
        print(f"An error occurred: {e}")
    
    finally:
        if s is not None:
            s.close()

        
if __name__ == "__main__":
    data = b"\x05\x00\x00\x00\x08" * 1
    #data = b"\x09\x00\x00\x00\x02\x00\x00\x00\x00" * 4    
    test_query(data)
```

stenzek:
I've implemented this in https://github.com/stenzek/duckstation/commit/02fbfae6a08556cf1ceaae096aa5cf5828240381, but with a number of changes:

1. The option is opt-out, i.e. by default an anonymous mapping gets created. This is to reduce the risk of leaks if DuckStation crashes, and can't clean up after itself. Or, if you are debugging DuckStation, and don't let it exit itself.
2. The crash handler makes a last-ditch effort to unlink the shared memory object before dumping the stack. This way if DS does crash, we won't leak ~100MB of memory each time.

Please note that as I've stated above, this is **not** safe for use when manipulating MIPS code externally. There is no page protection on the external mapping, so it will not invalidate JIT blocks, and you'll likely be executing stale code. It is **only** safe for manipulating/writing data.

PR #3225, opened 2024-06-16
Update Japanese translation

nishinji:
First Japanese language update in a year.
I escaped from my homework...
image


stenzek:
You'll likely need to update the ImGui glyph ranges as well, I can do that post-merge, or if you want to, it's this script: https://github.com/stenzek/duckstation/blob/master/src/duckstation-qt/translations/update_glyph_ranges.py

(otherwise, if there are any new glyphs in the Big Picture strings, they won't display)

nishinji:
I could't understand how to run it (installed python tho)
Can you do it later?

stenzek:
No problem. Thanks for your work :)

PR #3230, opened 2024-06-19
Fix binary_span_reader_writer.h (off-by-one error)

TheUbMunster:
Found what I believe to be an off-by-one error while debugging a PINE api user.

When stack trace is:

BinarySpanReader::Read(void* buf, unsigned __int64 size) `(the function I believe that contains the error)`
BinarySpanReader::ReadT<u32>()
BinarySpanReader::ReadU32() `<-- called by switch case no. MsgRead64: "const PhysicalMemoryAddress addr = rdbuf.ReadU32();"`
PINESocket::HandleCommand(PINEServer::IPCCommand command, BinarySpanReader rdbuf)
PINESocket::ProcessCommandsInBuffer()
...

and m_pos == 0 && size == 4 && m_buf.size() == 4 (large enough to contain a u32)
the comparison fails, even though I think it should succeed and pull out a u32, followed by moving m_pos forward 4 bytes.

Feel free to contact me if you require further details.

Also, If I'm wrong or making an incorrect assumption, please let me know!

### DIFF
--- a/src/common/binary_span_reader_writer.h
+++ b/src/common/binary_span_reader_writer.h
@@ -62,7 +62,7 @@ public:
 private:
   ALWAYS_INLINE bool Read(void* buf, size_t size)
   {
-    if ((m_pos + size) < m_buf.size()) [[likely]]
+    if ((m_pos + size) <= m_buf.size()) [[likely]]
     {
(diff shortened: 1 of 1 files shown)


stenzek:
Good catch, classic off by one when I inverted the condition.

PR #3238, opened 2024-06-30
 GL/ContextEGL: Fix red_size to blue_size 

Kpada:
Address an issue in the `OpenGLContextEGL::GetSurfaceTextureFormat()` function where the `blue_size` variable was incorrectly referenced as `red_size` in the condition checking.


stenzek:
I'd be surprised if this actually ever caused any issues, since (AFAIK) all 8bpc formats have at least 8 bits across all channels, but it's still incorrect regardless :)

Kpada:
Indeed, I meant a `possible` issue :)

PR #3243, opened 2024-07-06
fix: few bugs and improvements

Kpada:
Fix:
- access to a moved `path` vairable
- possible out-of-border access for `m_entries`

Improve:
- duplicated `Strncasecmp` checks
- `std::string` -> c-string -> `std::string` conversion
- attempt to move a `const` variable


stenzek:
Sounds very static analysis-ey. But still valid regardless, thanks :)

PR #3254, opened 2024-07-19
Qt/PAD: Enable/Disable Player LED for DualSense Controllers in Enhanced mode

dreamsyntax:
Port of:
https://github.com/PCSX2/pcsx2/pull/10864

Note, since the [Copy function here](https://github.com/PCSX2/pcsx2/pull/10864/files#diff-91056592066d19480eb6fd8b5387fb67d262b932a3c95b1d22362beac3712c57R333) does not exist in duckstation, this line was not added, everything else works just as on PCSX2.
I noticed RawInput was also not ported, but I didn't want to clutter this PR with both changes.

Reposting my description I originally posted in the PCSX2 PR below.

---- 
### Description of Changes
This change adds a toggle to enable/disable the Player LED on DualSense controllers.
Not to be confused with the primary LED. The Player LED is a non-customizable bright white array of white LEDs that indicate the player number. Image of before and after.
image

The UI change is added to the LEDSettingsDialog / controllerledsettingsdialog.ui
image

### Rationale behind Changes
* RAW mode does not have Player LEDs on.
* Turning on Enhanced Mode turns on the Player LED in addition to the primary LED.
* There is no option to disable the Player LEDs currently. You can only customize the primary LED.
* A user may want to completely disable all LEDs on their controller. Currently you can disable the Primary LED but not the Player LED. This is distracting/bright when playing in low lighting conditions.
* You can already tell which player is which due to the customized primary LED colors.

### Suggested Testing Steps
1. Connect DualSense controller. Note the player LED is off.
2. Turn on Enhanced DS4/DS5 mode under the DuckStation Controller Settings.
-- if on master --
3. Note the Player LED is ON. You can customize the Main LED but not the Player LED.
-- if on this branch --
3. Note the Player LED is still OFF.
4. Enter the LED Settings Dialog [controllerledsettingsdialog.ui] (The lightbulb icon) and utilize the new Checkbox. You should see all connected controllers enable/disable their player indicator LEDs. You can customize the Main LED per-controller as usual.
----

### DIFF
--- a/src/duckstation-qt/controllerledsettingsdialog.ui
+++ b/src/duckstation-qt/controllerledsettingsdialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>501</width>
-    <height>108</height>
+    <height>128</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -26,14 +26,14 @@
      </layout>
     </widget>
    </item>
-   <item row="0" column="1">
-    <widget class="QGroupBox" name="groupBox_2">
+   <item row="0" column="3">
+    <widget class="QGroupBox" name="groupBox_4">
      <property name="title">
-      <string>SDL-1 LED</string>
+      <string>SDL-3 LED</string>
      </property>
-     <layout class="QVBoxLayout" name="verticalLayout_3">
+     <layout class="QVBoxLayout" name="verticalLayout">
       <item>
-       <widget class="ColorPickerButton" name="SDL1LED"/>
+       <widget class="ColorPickerButton" name="SDL3LED"/>
       </item>
      </layout>
     </widget>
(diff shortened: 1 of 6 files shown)


src/duckstation-qt/controllerledsettingsdialog.ui
```diff
@@ -1,83 +1,112 @@
...
+			<rect>
+				<x>0</x>
+				<y>0</y>
+				<width>501</width>
+				<height>128</height>
+			</rect>
+		</property>
+		<property name="windowTitle">
+			<string>Controller LED Settings</string>
+		</property>
```
dreamsyntax, line 15: I'm not really clear what's going on with the spacing, as it doesn't look like this when pulled locally. If this is a blocker let me know.

stenzek: Did you edit it with Qt designer? It looks like it's been changed from 1 space indent to tabs.

dreamsyntax: Ah, yeah that was the issue. Visual Studio was reformatting it after viewing/saving for me.

PR #3255, opened 2024-07-20
Update SDL game controller database

Mrlinkwii:
was last updated in 2021 , not much has changed   since , it brings the SDL game controller database up to whats  the latest


stenzek:
Actually last updated on June 23 this year :)

PR #3286, opened 2024-09-01
DiscordRPC: Add Rcheevos Icon Support

kamfretoz:
This PR adds Rcheevos icon support to the Discord's Rich Presence

Preview:
Screenshot_20240901_211909

### DIFF
--- a/src/core/system.cpp
+++ b/src/core/system.cpp
@@ -5892,6 +5892,8 @@ void System::UpdateRichPresence(bool update_session_time)
     const auto lock = Achievements::GetLock();
     state_string = StringUtil::Ellipsise(Achievements::GetRichPresenceString(), 128);
     rp.state = state_string.c_str();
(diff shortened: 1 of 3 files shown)


src/core/system.cpp
```diff
@@ -5892,6 +5892,7 @@ void System::UpdateRichPresence(bool update_session_time)
     const auto lock = Achievements::GetLock();
     state_string = StringUtil::Ellipsise(Achievements::GetRichPresenceString(), 128);
     rp.state = state_string.c_str();
+    rp.largeImageKey = Achievements::GetGameIconURL().c_str();
```
stenzek, line 5895:
```suggestion
    if (const std::string& icon_url = Achievements::GetGameIconURL(); !icon_url.empty())
      rp.largeImageKey = icon_url.c_str();
```

Icon URL may not exist.

kamfretoz: Fixed.

PR #3287, opened 2024-09-02
GPU: Add "Force Frame Timings" option

spacepaw:
## Description
I've revamped the "Force NTSC Timings" option to enable users to enforce both NTSC and PAL timings. The option has been renamed to "Force Frame Timings" and I've replaced the toggle with a drop-down menu.

## Reason for the PR
Many homebrew PS1 projects rely on Hitmod as their music player. In earlier versions of Hitmod, music doesn’t play at the correct tempo at 60fps. This new option allows NTSC productions to run at 50fps, ensuring proper music playback.


stenzek:
Thanks! Few minor points WRT localization, I'll sort those out post-merge.

Will probably rename the setting to "Force Video Timing" rather than "Frame", since "Frame" is ambiguous between video and audio.

Also need to merge the disable interlacing and deinterlacing settings to make room for this one, otherwise the 768p'ers will complain. But that's a me problem for later.

PR #3289, opened 2024-09-04
GameDB: Fix MGS Premium Package

bslenul:
Renamed discs according to Redump, added the alternative serials and removed the incorrect entries where the game was considered as a single disc.

### DIFF
--- a/data/resources/gamedb.yaml
+++ b/data/resources/gamedb.yaml
@@ -99027,9 +99027,9 @@ PAPX-90044:
   traits:
     - ForcePGXPCPUMode
 SCPS-45317:
-  name: "Metal Gear Solid (Asia) (Disc 1)"
+  name: "Metal Gear Solid (Japan, Asia) (Disc 1) (Ichi) (Premium Package)"
   discSet:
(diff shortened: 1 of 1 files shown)


data/resources/gamedb.yaml
```diff
@@ -99027,9 +99027,9 @@ PAPX-90044:
   traits:
     - ForcePGXPCPUMode
 SCPS-45317:
-  name: "Metal Gear Solid (Asia) (Disc 1)"
+  name: "Metal Gear Solid (Japan, Asia) (Disc 1) (Ichi) (Premium Package)"
   discSet:
-    name: "Metal Gear Solid (Asia)"
+    name: "Metal Gear Solid (Japan, Asia) (Premium Package)"
```
stenzek, line 99032: Hopefully this one wasn't actually being detected before, because otherwise it's going to change memory card names.

stenzek: Eh, don't worry about it. It's not like it deletes old cards, users would just have to rename them.

PR #3290, opened 2024-09-05
Add *.psx as a supported PS-X EXE file type

spacepaw:
## Description
I've added *.psx as a supported PS-X EXE file type. The reason for this addition is that many PS1 demos and intros available on Pouet and Demozoo are in .psx format.


stenzek:
Seems fine, thanks!

PR #3294, opened 2024-09-07
Fixed issue with cheats import failing on DS format cheats

PugsyMAME:
Reverted part of cheats.cpp back to earlier code. The new code causes the import of DuckStation format cheats to fail unnecessarily.  Eg. Importing this will fail: 

#Show sides of screen that are blacked out
A706E00A 0C012400
A706DFD2 0C012400

### DIFF
--- a/src/core/cheats.cpp
+++ b/src/core/cheats.cpp
@@ -661,10 +661,8 @@ bool CheatList::LoadFromString(const std::string& str, Format format)
     return LoadFromPCSXRString(str);
   else if (format == Format::Libretro)
     return LoadFromLibretroString(str);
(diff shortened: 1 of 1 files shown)


stenzek:
Looks like detection is the issue here. Can you please check if it is fixed after https://github.com/stenzek/duckstation/commit/9d5a84541f0b0f0c726948d22b36e8532e4a0ac4?

stenzek:
Confirmed fixed in https://github.com/stenzek/duckstation/commit/9d5a84541f0b0f0c726948d22b36e8532e4a0ac4#commitcomment-146404299.

PR #3297, opened 2024-09-14
Quality of Life changes to breakpoints

PugsyMAME:
1. Made it so that breakpoints are checked for a match on 28 bits rather than the full 32 bits. It's much simpler to use 12345 rather than 80012345 and risk getting the number of zeros wrong and also handles the mirror code/memory accesses using 00??????, 80?????? & A0??????. 
2. Added bp.hit_count update to execution breakpoints.
3. Changed the Hit breakpoint message at the bottom of the debugger screen to include the hit count [see notes later].
4. Added bp.enabled check to the execution breakpoint (it uses it in the r/w breakpoints code already) [see notes later]. 

Notes: 
3. I've added the hit count to the message as it's quicker to spot but it's also a partial workaround as the Hit Count in the Breakpoints tab doesn't seem to get refreshed. 
4. I thought the checkbox in the Breakpoints tab would set the bp.enabled but it seems it has no affect on any type of breakpoint. The only way to stop a breakpoint is to delete it.

### DIFF
--- a/src/core/cpu_core.cpp
+++ b/src/core/cpu_core.cpp
@@ -2113,10 +2113,13 @@ const char* CPU::GetBreakpointTypeName(BreakpointType type)
 
 bool CPU::HasBreakpointAtAddress(BreakpointType type, VirtualMemoryAddress address)
 {
-  for (const Breakpoint& bp : GetBreakpointList(type))
+  for (Breakpoint& bp : GetBreakpointList(type))
   {
-    if (bp.address == address)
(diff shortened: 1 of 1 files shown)


stenzek:
>  It's much simpler to use 12345 rather than 80012345

This is _technically_ not correct, but I agree, it's a QOL improvement, and the only place where it wouldn't "work" would be the scratchpad, and that's not executable anyway. Made the mistake of forgetting BPs weren't masked myself too.

PR #3301, opened 2024-09-23
gitgnore : add /.cache/ folder

Mrlinkwii:
self-explanatory


stenzek:
What is this for? I don't have it in any of my setups.

PR #3319, opened 2024-10-19

andercard0:
Confirmado linhas movidas;
Adicionado linhas referente aos 2 novos controles.

@stenzek Could you please check if I mess something up... I always test before PR but the translation is not being shown on emulator as below:

Controle

and here the title is correct too but not the buttons:

image


stenzek:
Sorry, that was my error - forgot to change the context on the new strings. Should be fixed once you lupdate and confirm the new context.

PR #3328, opened 2024-11-10
Cheat bugfix and new A8 cheat type

PugsyMAME:
* Added needed g_emu_thread->reloadCheats calls after the reloadList() calls on entering a new code or importing new codes.

Without it I had to import new codes and then manually edit one of them for it to show up in the cheat counts and possibly even work

* Added cheat type A8 which is the byte equivalent of the A7 cheat type as there will be a need of this type.

### DIFF
--- a/src/duckstation-qt/gamecheatsettingswidget.cpp
+++ b/src/duckstation-qt/gamecheatsettingswidget.cpp
@@ -529,7 +529,7 @@ void GameCheatSettingsWidget::importCodes(const std::string& file_contents)
   }
 
   reloadList();
-  g_emu_thread->reloadCheats(true, true, false, true);
+  g_emu_thread->reloadCheats(true, false, false, true);
 }
 
 void GameCheatSettingsWidget::newCode()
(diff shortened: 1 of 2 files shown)


src/duckstation-qt/gamecheatsettingswidget.cpp
```diff
@@ -529,6 +529,7 @@ void GameCheatSettingsWidget::importCodes(const std::string& file_contents)
   }
 
   reloadList();
+  g_emu_thread->reloadCheats(true, true, false, true);
```
stenzek, line 532: I'm not sure about this. It shouldn't need to reload cheats, since the one you just imported should not be enabled yet.

Once you click the checkbox, it should trigger a reload.

PugsyMAME: You mentioned on Discord what the bools should be and I can confirm that true, false,false ,true works for both lines.

  g_emu_thread->reloadCheats(true, false, false, true);

I've updated it on my master branch to reflect this.

PR #3334, opened 2024-11-27
Debugger/MemoryScanner: Add 'Freeze Selected'

dreamsyntax:
* Adds 'Freeze Selected Entries' which upon multiple watch entries selected, will freeze/unfreeze them to the opposite of the first entry's state.
* Change "Remove Selected Entries from Watch List" -> "Remove Selected Entries"

This functionality has been useful for me. If you dislike this approach please don't hesitate to let me know and I'll change it. I figure it would be nice to have this feature.

Screenshots

### Before
image

### After
image

### DIFF
--- a/src/core/memory_scanner.cpp
+++ b/src/core/memory_scanner.cpp
@@ -340,6 +340,14 @@ bool MemoryWatchList::AddEntry(std::string description, u32 address, MemoryAcces
   return true;
 }
 
+bool MemoryWatchList::GetEntryFreeze(u32 index) const
+{
+  if (index >= m_entries.size())
+    return false;
+
+  return m_entries[index].freeze;
+}
+
 void MemoryWatchList::RemoveEntry(u32 index)
 {
   if (index >= m_entries.size())
--- a/src/core/memory_scanner.h
+++ b/src/core/memory_scanner.h
@@ -112,6 +112,8 @@ public:
   u32 GetEntryCount() const { return static_cast<u32>(m_entries.size()); }
 
   bool AddEntry(std::string description, u32 address, MemoryAccessSize size, bool is_signed, bool freeze);
+  bool GetEntryFreeze(u32 index) const;
+
   void RemoveEntry(u32 index);
   bool RemoveEntryByDescription(const char* description);
   bool RemoveEntryByAddress(u32 address);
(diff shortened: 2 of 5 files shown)


src/core/memory_scanner.h
```diff
@@ -112,6 +112,8 @@ class MemoryWatchList
   u32 GetEntryCount() const { return static_cast<u32>(m_entries.size()); }
 
   bool AddEntry(std::string description, u32 address, MemoryAccessSize size, bool is_signed, bool freeze);
+  bool GetEntryFreeze(u32 index);
```
stenzek, line 115:
```suggestion
  bool GetEntryFreeze(u32 index) const;
```

src/core/memory_scanner.cpp
```diff
@@ -340,6 +340,14 @@ bool MemoryWatchList::AddEntry(std::string description, u32 address, MemoryAcces
   return true;
 }
 
+bool MemoryWatchList::GetEntryFreeze(u32 index)
```
stenzek, line 343:
```suggestion
bool MemoryWatchList::GetEntryFreeze(u32 index) const
```

src/duckstation-qt/memoryscannerwindow.cpp
```diff
@@ -330,6 +332,22 @@ void MemoryScannerWindow::addManualWatchAddressClicked()
...
 }
 
+void MemoryScannerWindow::freezeWatchClicked()
+{
+  const int indexFirst = getSelectedWatchIndexFirst();
+  const int indexLast = getSelectedWatchIndexLast();
+  if (indexFirst < 0)
+    return;
+
+  const bool freeze = m_watch.GetEntryFreeze(indexFirst);
```
stenzek, line 342: I'm a little confused on this. It's effectively a toggle based on the first selected item.

Wouldn't you either way to implement it as a proper toggle, i.e. flip each entry depending on its state, or, just set them all to enabled?

stenzek (commented):
Otherwise seems fine.

src/duckstation-qt/memoryscannerwindow.cpp
dreamsyntax: > I'm a little confused on this. It's effectively a toggle based on the first selected item.
> 
> Wouldn't you either way to implement it as a proper toggle, i.e. flip each entry depending on its state, or, just set them all to enabled?

Yes, its a toggle based on the first item since the real use case flow is usually either to mass enable or mass disable the freeze.
In real use you'd pretty much never want to independently flip with a mass selection, so rather than have dedicated 'freeze' and 'unfreeze' buttons, I use the first selected's state.
At the worst case you'd just have to press the button 3 times to sync your selection.

stenzek: Okay, if that makes sense for you, then sure.

I'm not using the feature, so don't have a particularly strong opinion either way.

PR #3361, opened 2025-02-06
update Chinese translation.

zkdpower:
just Translated User Interface.

please add font "**污 冗 脚 钮 省 奖 杯 阴 穿 客 舍 五**" for font libary of "big picture mode"'.

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_zh-CN.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_zh-CN.ts
@@ -4,9 +4,8 @@
 <context>
     <name>AboutDialog</name>
     <message>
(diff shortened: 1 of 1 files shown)


crashGG:
一个勘误："Round Upscaled Texture Coordinates" 为升格纹理坐标后取整，round意思是舍入取整，来源于函数 round. 升格多用于图像领域整数放大。
An errata: "Round Upscaled Texture Coordinates" is 升格纹理坐标后取整 , where **round** comes from the meaning of "round" in the function.

stenzek:
FYI I don't see any glyphs that aren't currently in the range for big picture mode.

PR #3377, opened 2025-02-23

zkdpower:
the following string need add for translation.

the icon of "ISO Browser" and "Free Camera" should be displaced to a default duck icon.


stenzek:
> the icon of "ISO Browser" and "Free Camera" should be displaced to a default duck icon.

These windows already have specific icons set.

PR #3382, opened 2025-03-01
GameDB: Kagero II & III

6lackmag3:
This removes Analog Controller support for specified games as they only support vibration.


stenzek:
Manually merged in https://github.com/stenzek/duckstation/commit/b8d86a18da851b8da444eb72f9fdc947bd2f379d due to extra spaces being added and failing the check.

6lackmag3:
Thanks for fixing my mistake!

PR #3388, opened 2025-03-13

zkdpower:
please add "**粉，曜，繁**" font for BIG-picture UI.


zkdpower:
**"Formatting the memory card will destroy all saves, and they will not be recoverable. The memory card which will be formatted is located at '%1'." this string seems to be prombelm，It is unable to display the memory card file address.**

stenzek:
That string works fine/as expected here.

PR #3423, opened 2025-05-23
Pin action-automatic-releases to latest commit

Pesa:
GitHub Actions supply chain attacks are sadly a thing ([recent example](https://unit42.paloaltonetworks.com/github-actions-supply-chain-attack/)) and this action is in a privileged position where it could potentially do a lot of damage.

The bigger issue is that this action has been unmaintained for >3 years, but addressing that will require much larger changes to the release workflow and I wasn't sure how to proceed.

Let me know if I should pin the flatpak actions too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Create Preview Release
         if: github.ref == 'refs/heads/master'
-        uses: "marvinpinto/action-automatic-releases@latest"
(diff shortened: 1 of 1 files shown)


stenzek:
Whoops, guess I missed this one when I went through and pinned them all a while back..

> Let me know if I should pin the flatpak actions too.

I originally had it pinned, but the flatpak folks said to change it to a non-pinned version. Personally I don't care much for the flatpak, and the only secrets in the repo are used for the flatpak, so... whatever.

PR #3427, opened 2025-05-26
Qt: Further polish the layout of some widgets

Pesa:
A few more minor fixes to the layout of several settings widgets and dialogs to improve the appearance with the native theme on macOS. As usual, Qt designer reorders the rows and thus makes the diff larger than it needs to be. The actual changes are as follows:

* Remove redundant QHBoxLayout in audio, bios, and advanced settings. It was causing weird spacing/misalignment issues despite all margins being zero. Probably a Qt bug.
* Prevent multi-line text in labels from being cut off in `coverdownloaddialog.ui`. This is the same issue I fixed earlier for the audio stretch settings, but I found a better solution so I also applied it to `audiostretchsettingsdialog.ui`. This should be more robust and avoid the issue even if the user resizes the dialog or uses a different (larger) font size.

### DIFF
--- a/src/duckstation-qt/advancedsettingswidget.ui
+++ b/src/duckstation-qt/advancedsettingswidget.ui
@@ -28,7 +28,7 @@
      <property name="title">
       <string>Logging</string>
      </property>
-     <layout class="QGridLayout" name="gridLayout_3" columnstretch="0,1">
+     <layout class="QGridLayout" name="gridLayout_3">
       <item row="0" column="0">
        <widget class="QLabel" name="label">
         <property name="text">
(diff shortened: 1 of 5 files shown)


Pesa:
> It was causing weird spacing/misalignment issues despite all margins being zero. Probably a Qt bug.

Yep, looks like a very old Qt bug that they never bothered to fix...
https://bugreports.qt.io/browse/QTBUG-126045
https://bugreports.qt.io/browse/QTBUG-63867
https://forum.qt.io/topic/62374/unwanted-vertical-padding-with-qcombobox-in-nested-layout

PR #3429, opened 2025-05-27
GameDB: Various fixes to GTA games

CRASHARKI:
- None of the games support analog control.
- Only the London games support vibration. 
- Changed the GTA (Europe) discs to use the correct London game (Mission Pack instead of Standalone).
- Various "available languages" corrections.


stenzek:
In the future for these games that support vibration but not analog, mode, could you please leave AnalogController in place, and instead add:

```yaml
  traits:
    - DisableAutoAnalogMode # Analog mode is not supported.
```

Thanks.

PR #3432, opened 2025-05-30
GameDB: Disable upscaling and filtering in Constructor

CRASHARKI:


crashGG:
Some 2D games can get a visual boost from mmpx sprite filtering, which requires 2x scaling.

PR #3433, opened 2025-05-31
Avoid deprecated `std::memory_order_consume`

Pesa:
`std::memory_order_consume` was deprecated in C++26, and even before then, I don't know of any compiler that actually implements it differently from `memory_order_acquire`.


stenzek:
This one could just be removed completely, there's no corresponding release.

PR #3434, opened 2025-06-01
GameDB: Two fixes for `Alone in the Dark - The New Nightmare`

Pesa:
* The game hangs immediately after the character selection screen with read speedup = maximum. 150k cycles of delay fixes that. Not sure if this is worth it or we should just add the `DisableCDROMReadSpeedup` trait instead.
* `displayLineEndOffset` prevents some text at the bottom from being cut off. See screenshot. We're still debating on discord whether to use `displayCropMode: Borders` instead, hence the draft status.

overscan


Pesa:
I decided to go with `displayCropMode: Borders` as suggested on discord since it also removes a tiny black bar on the left of the screen.

PR #3437, opened 2025-06-05
README: Fix typo

Pesa:
Qt 6.8 bumped the minimum target platform to Monterey. DuckStation has been using 6.9.x for some time.

Also replace `x64` with `x86_64`, the latter is more common on Apple platforms.


stenzek:
In my experience it tends to still work despite that. I haven't changed the min version in duckstation itself.

PR #3442, opened 2025-06-11
Qt: Implement reset play time for disc sets

Pesa:
A few things I'm not sure about:

* ~Should I use `entry->disc_set_name` or `entry->title`? is there a difference if the entry is a discset?~ Ended up using `entry->path` because `disc_set_name` is empty for the actual `DiscSet` list entry.
* I don't fully understand the locking (and the use of a recursive mutex doesn't make it any easier). It seems that the whole thing is already executed with the lock held, starting from `MainWindow::onGameListEntryContextMenuRequested`, so maybe I don't need to lock again in `clearGameListEntryPlayTime()`? Assuming that the triggered menu action is executed synchronously inside `menu.exec()`...
* I kept changes to a minimum but, if desired, repeated invocations of `GameList::ClearPlayedTimeForSerial` could be optimized by passing a list of serials, so that we iterate `s_entries` only once when resetting multiple entries.

### DIFF
--- a/src/duckstation-qt/mainwindow.cpp
+++ b/src/duckstation-qt/mainwindow.cpp
@@ -1047,7 +1047,7 @@ void MainWindow::onCheatsMenuAboutToShow()
 const GameList::Entry* MainWindow::resolveDiscSetEntry(const GameList::Entry* entry,
                                                        std::unique_lock<std::recursive_mutex>& lock)
 {
-  if (!entry || entry->type != GameList::EntryType::DiscSet)
+  if (!entry || !entry->IsDiscSet())
     return entry;
 
   // disc set... need to figure out the disc we want
@@ -1523,14 +1523,6 @@ void MainWindow::onGameListEntryContextMenuRequested(const QPoint& point)
           switchToEmulationView();
         });
       }
-
-      menu.addSeparator();
-
-      connect(menu.addAction(tr("Exclude From List")), &QAction::triggered,
-              [this, entry]() { getSettingsWindow()->getGameListSettingsWidget()->addExcludedPath(entry->path); });
-
-      connect(menu.addAction(tr("Reset Play Time")), &QAction::triggered,
-              [this, entry]() { clearGameListEntryPlayTime(entry); });
     }
     else
     {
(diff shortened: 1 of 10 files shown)


src/duckstation-qt/mainwindow.cpp
```diff
@@ -1555,12 +1547,15 @@ void MainWindow::onGameListEntryContextMenuRequested(const QPoint& point)
...
       connect(menu.addAction(tr("Select Disc")), &QAction::triggered, this, &MainWindow::onGameListEntryActivated);
+    }
 
-      menu.addSeparator();
+    menu.addSeparator();
 
-      connect(menu.addAction(tr("Exclude From List")), &QAction::triggered,
-              [this, entry]() { getSettingsWindow()->getGameListSettingsWidget()->addExcludedPath(entry->path); });
-    }
+    connect(menu.addAction(tr("Exclude From List")), &QAction::triggered,
```
stenzek, line 1554: This one's a bit iffy - disc sets don't have a path, so excluding them doesn't really make a lot of sense.

It'd probably "work", but to the user, a name in an excluded path list might be confusing.

src/duckstation-qt/mainwindow.cpp
```diff
@@ -1628,7 +1623,17 @@ void MainWindow::clearGameListEntryPlayTime(const GameList::Entry* entry)
     return;
   }
 
-  GameList::ClearPlayedTimeForSerial(entry->serial);
+  if (!entry->IsDiscSet())
```
stenzek: It would be better to move this logic to `GameList`, that way it can be used by both Big Picture and the Qt UI.

Something like `void GameList::ClearPlayedTimeForEntry(const Entry* entry)`.

stenzek:
> I don't fully understand the locking (and the use of a recursive mutex doesn't make it any easier). It seems that the whole thing is already executed with the lock held, starting from MainWindow::onGameListEntryContextMenuRequested, so maybe I don't need to lock again in clearGameListEntryPlayTime()? Assuming that the triggered menu action is executed synchronously inside menu.exec()...

It's a bit janky, yeah. `exec()` is blocking, so it won't return until the menu is closed. This keeps `entry` alive, since it is captured by reference in the callbacks. The recursive mutex is needed in case any of those actions mutate game list state.

The downside is that scanning will effectively get halted while a context menu is open. We'd have to create copies of the entry's fields for the actions/callbacks to avoid that, and is actually what I've done in my other duckstation-derived host.. should probably backport it.

src/duckstation-qt/mainwindow.cpp
Pesa: This was already in both branches of the if/then, I just moved it below to avoid repetition. Are you saying it should not be there for disc sets?

stenzek: Eh, if it's already there, whatever.

Pesa:
It's tested and working now.

PR #3443, opened 2025-06-12
enhance MMPX algorithm, fix some minor graphics defects.

crashGG:
Enhance MMPX, Fixed some defects in the original algorithm by morphological analysis

MediEvil (USA)
Pay attention to the bubbles in  the right eye socket
master:
MediEvil (USA) master
pr:
MediEvil (USA) pr

Rockman X6 （Japan）
bubbles on the body and bomb
master:
Rockman X6 master
pr:
Rockman X6 pr

Tekken 2 (USA)
bubbles on the letters and Blurred grid on integrating sphere
master:
Tekken 2 (USA)master
pr:
Tekken 2 pr

### DIFF
--- a/src/core/gpu_hw_shadergen.cpp
+++ b/src/core/gpu_hw_shadergen.cpp
@@ -742,7 +742,7 @@ void FilteredSampleFromVRAM(TEXPAGE_VALUE texpage, float2 coords, float4 uv_limi
     ss << R"(
 uint luma(uint C) {
     uint alpha = (C & 0xFF000000u) >> 24;
-    return (((C & 0x00FF0000u) >> 16) + ((C & 0x0000FF00u) >> 8) + (C & 0x000000FFu) + 1u) * (256u - alpha);
+    return (((C & 0x00FF0000u) >> 16) + ((C & 0x0000FF00u) >> 8) + (C & 0x000000FFu)) + (255u - alpha) * 3;
 }
 
 bool all_eq2(uint B, uint A0, uint A1) {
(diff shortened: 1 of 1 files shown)


stenzek:
This probably should be a separate mode, the shader is much heavier.

What is the source for the algorithm? Should name it after that.

stenzek:
Okay. I'll merge this and then split it out to original and enhanced variants.

> A practical experience is that I hardly observed an increase in hardware usage on ppsspp compared to the original MMPX.

The approach is different in DuckStation. Because of how the PSX hardware works and the fact that multiple textures can co-exist in the same page, filtering is applied at draw time by performing the upscaling for **every pixel that is rendered**, interpolating between or selecting between the upscaled texel values based on the fractional texture coordinates.

So, the performance hit will be larger.

stenzek:
FWIW, I'm seeing an increase from 1.33ms/frame to 1.44ms/frame in Megaman X6 on an AMD 7900GRE, so only ~8% overhead, not bad considering it definitely looks better :).

crashGG:
Report an issue. 
Currently, there will be edge display issues of alpha-semi-transparent textures after scale filtering, including using xBR, EPX, MMPX
Valkyrie Profile (Japan) (Disc 1)
This scene is 1 minute after the new game starts
Valkyrie Profile (Japan) (Disc 1) 2025-06-13-11-02-21

Ordinary textures also appear when fading in and out.
MediEvil

stenzek:
From a very quick glance, VP looks to just be an issue with how it's rendering - bilinear also has artifacts on the edges.

They're drawing a black blended background, and then blending the sprite on top. Same UVs, but different CLUTs/palettes, one gets alpha culled, the other does not.

PR #3447, opened 2025-06-14
Qt: Avoid showing "0 minutes" of played time

Pesa:
Kind of an edge case, but seeing "0 minutes" may be confusing for the user.


stenzek:
Probably should just do this in GameList instead.

PR #3452, opened 2025-06-17
Achievements: Use `gethostuuid()` as machine key on macOS

Pesa:


src/core/achievements.cpp
```diff
@@ -3741,12 +3740,21 @@ static TinyString GetLoginEncryptionMachineKey()
 
   ret.resize(machine_guid_length);
   RegCloseKey(hKey);
-#elif !defined(__ANDROID__)
```
Pesa, line 3744: this whole function is inside a `#ifndef __ANDROID__` block, so the extra condition here was redundant.

src/core/achievements.cpp
```diff
@@ -3741,12 +3740,21 @@ static TinyString GetLoginEncryptionMachineKey()
...
+#else
+#if defined(__linux__)
   // use /etc/machine-id on Linux
   std::optional<std::string> machine_id = FileSystem::ReadFileToString("/etc/machine-id");
   if (machine_id.has_value())
     ret = std::string_view(machine_id.value());
+#elif defined(__APPLE__)
+  // use gethostuuid(2) on macOS
+  const struct timespec ts = { .tv_sec = 2, .tv_nsec = 0 };
+  uuid_t uuid = {};
```
Pesa: `uuid_t` is essentially an `unsigned char[16]`

src/core/achievements.cpp
```diff
@@ -3741,12 +3740,21 @@ static TinyString GetLoginEncryptionMachineKey()
...
-#ifdef __linux__
+#else
+#if defined(__linux__)
   // use /etc/machine-id on Linux
   std::optional<std::string> machine_id = FileSystem::ReadFileToString("/etc/machine-id");
   if (machine_id.has_value())
     ret = std::string_view(machine_id.value());
+#elif defined(__APPLE__)
+  // use gethostuuid(2) on macOS
+  const struct timespec ts = { .tv_sec = 2, .tv_nsec = 0 };
```
stenzek: Is 2 seconds wait time reasonable? If this is actually an expensive operation, I'd be concerned about "random" slowdowns to due to machine load causing a different key to be used and invalidating a user's login.

src/core/achievements.cpp
```diff
@@ -3741,12 +3740,21 @@ static TinyString GetLoginEncryptionMachineKey()
...
+  uuid_t uuid = {};
+  if (gethostuuid(uuid, &ts) == 0)
+  {
+    for (auto c : uuid)
+      ret.append_format("{:02x}", c);
```
stenzek:
```suggestion
    ret.append_hex(uuid, sizeof(uuid), false);
```

Assuming it's typedef'ed to an array.

src/core/achievements.cpp
Pesa: yeah good point. TBH I don't know why there's a timeout for a function call like this. Maybe it does some IPC? The call was instantaneous for me, even on a 2019 macbook. The timeout can be set to infinite, that could cause other issues if the call blocks forever but I don't know if that can even happen in practice.

stenzek: I'd just make it infinite.

PR #3454, opened 2025-06-17
Flatpak: Grant access to Gamescope socket for Steam Deck

BlueInterlude:
* See flathub/org.DolphinEmu.dolphin-emu#178
* Bug only occurs on the Steam Deck OLED when using the Vulkan renderer.

To replicate the error, on a Steam Deck OLED, install the Gamescope Flatpak, set renderer to Vulkan, launch a ROM in Game Mode through the DuckStation GUI, see error. Gamescope Flatpak is primarily used for HDR in other Flatpaks (Dolphin, PrimeHack, Heroic, etc.).

image

### DIFF
--- a/scripts/packaging/flatpak/org.duckstation.DuckStation.yaml
+++ b/scripts/packaging/flatpak/org.duckstation.DuckStation.yaml
@@ -19,6 +19,8 @@ finish-args:
   - "--talk-name=org.freedesktop.ScreenSaver"
   - "--socket=wayland"
   - "--socket=fallback-x11"
+  # Required for Gamescope on Steam Deck OLED when using the Vulkan renderer
+  - --filesystem=xdg-run/gamescope-0:ro
 
 modules:
   # Dependencies.


BlueInterlude:
> Thanks for the PR. I'd be hesitant about adding any permission since flathub displays scary warnings for filesystem access, making it sound like my application is malicious.
> 
> Additionally, their silly system requires manual approval of any changes, and the last time it changed, they approved the aarch64 variant but not x86. So I had to push a dummy commit and another release, spamming all users with a notification just because of this pointless process. Much larger app stores such as google play don't require such rubbish.
> 
> I'm considering dropping the flatpak for this reason and others, but a middle ground where the CI builds bundles but they're not published on flathub is an option. In which case the added permission is a non issue, and I can just request full filesystem r/o access which solves the portal paths constantly breaking problem too.
> 
> Generally speaking, I'm not a huge fan of workarounds for one specific piece of hardware that I don't own and can't debug myself. This sounds like something that should be covered in flatpak itself, not applications.
> 
> Alternatively, why can't users just manually add the permission themselves? Or, of course, use the appimage.

No problem, makes sense to me.

Users can add the permission themselves, but in the case you're installing the Gamescope Flatpak to get HDR working for the Dolphin emulator and then open DuckStation days or weeks later and see this error, putting 2 and 2 together and realizing that it's a permissions issue might not be obvious for most. 

I agree though, some sort of better error message from the whole Flatpak system than the one I posted in the original comment would be much more helpful for debugging and realizing what's actually happening. 

Can leave this PR open in case you decide to go the CI route though or close, either way works for me!

PR #3455, opened 2025-06-18
Two fixes for `AchievementLoginDialog`

Pesa:
Before:
before

After:
after

### DIFF
--- a/src/duckstation-qt/achievementlogindialog.ui
+++ b/src/duckstation-qt/achievementlogindialog.ui
@@ -3,26 +3,20 @@
  <class>AchievementLoginDialog</class>
  <widget class="QDialog" name="AchievementLoginDialog">
   <property name="windowModality">
-   <enum>Qt::WindowModal</enum>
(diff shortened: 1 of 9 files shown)


src/duckstation-qt/achievementlogindialog.ui
```diff
@@ -40,24 +34,17 @@
...
       </widget>
      </item>
      <item>
       <widget class="QLabel" name="label_2">
-       <property name="font">
-        <font>
-         <pointsize>14</pointsize>
-         <weight>50</weight>
-         <bold>false</bold>
-        </font>
```
Pesa: This is not working on macOS (has no effect) so I've replaced it with Markdown text. Dunno if there's a better way... maybe the font size can be set programmatically? Also I have no idea how this looks on Windows (it's fine on Linux and macOS)

stenzek: The current setup works fine on Windows.. might be a uic issue and need to manually set the font.

PR #3464, opened 2025-06-25
GameDB: Disable read speedup in Spyro 2 & 3

Pesa:
Reported on discord.


6lackmag3:
Why disable completely? No issues here with 2x (Quad Speed).

Pesa:
I don't have the game to test.

Maybe it's a combination of read+seek speedup that's causing the issue. What's your seek speedup set to?

6lackmag3:
I'm using combination of cdr=2x cds=10x globally and faced no issues so far. You may have shorter loading times with cdr>2x in some games but it's a hit and miss and usually brings more troubles than benefits.

Pesa:
Are you talking about spyro 2 or 3 or both? PAL or NTSC?

Btw, stenzek explained to me that "maximum" speedup typically ends up being more compatible than using a high multiplier.

6lackmag3:
@Pesa ok, I've tested all three Spyro games, PAL and NTCS. The results are:
Spyro 1 - both verisons working fine with maximum CD read/seek speed up
Spyro 2 PAL, cdr 2x - audio issues during long cutscenes
Spyro 2 NTSC, cdr > 2x - audio issues in long cutscenes
Spyro 3 PAL, cdr = 2x - audio issues and occasional infinite loading screens 
Spyro 3 NTSC, cdr > 2x - infinite loading screens ("Loading game failed: no response")

So in terms of stability it's fair to disable CD read speadup for Spyro 2 and 3.

stenzek:
There's definitely games which need proper seek timing. But it's rare, yes. I listed a couple in https://github.com/stenzek/duckstation/wiki/Difficult-to-Emulate-Games (search "seek")

6lackmag3:
Yep, it's shaking on Windows. That's how I noticed it, actually.

> Not a bug, game is doing it deliberately if I remember correctly.
> 
> As long as there's no visual flickering.

stenzek:
Look fine to me. I see the resolution change, but the image stays static.

If you're using All Borders crop, then yes, it'll break. But that's expected for how that mode behaves.

PR #3475, opened 2025-07-06
Qt: Set default button in settings and ISO browser

Pesa:
Must have missed these in #3457

### DIFF
--- a/src/duckstation-qt/controllersettingswindow.cpp
+++ b/src/duckstation-qt/controllersettingswindow.cpp
@@ -27,10 +27,9 @@ ControllerSettingsWindow::ControllerSettingsWindow(INISettingsInterface* game_si
   : QWidget(parent), m_editing_settings_interface(game_sif), m_editing_input_profiles(edit_profiles)
 {
   m_ui.setupUi(this);
+  m_ui.buttonBox->button(QDialogButtonBox::Close)->setDefault(true);
 
   setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
-
-  m_ui.settingsCategory->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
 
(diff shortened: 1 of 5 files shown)


src/duckstation-qt/controllersettingswindow.cpp
```diff
@@ -27,11 +27,10 @@ ControllerSettingsWindow::ControllerSettingsWindow(INISettingsInterface* game_si
   : QWidget(parent), m_editing_settings_interface(game_sif), m_editing_input_profiles(edit_profiles)
 {
   m_ui.setupUi(this);
+  m_ui.buttonBox->button(QDialogButtonBox::Close)->setDefault(true);
 
   setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
 
-  m_ui.settingsCategory->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
```
Pesa, line 33: moved to the `.ui` file

src/duckstation-qt/isobrowserwindow.ui
```diff
@@ -133,6 +133,9 @@
        <property name="text">
         <string>Close</string>
```
Pesa, line 134: btw, a close button for the ISO browser feels unnecessary IMHO. The same goes for the cover downloader. Both are "tools" windows that feel somewhat independent from the main UI. The memcard editor, which is also standalone, already lacks a close button.

I may be influenced by OS-specific UI paradigms, as close (or OK) buttons are uncommon on macOS, except on modal dialogs. Maybe it's different on Windows...

stenzek: Yeah, I think there is a difference between the Windows and MacOS HIG here - as far as I'm aware, most dialogs on Windows will have some sort of action button that closes the dialog. Whether that's OK/Cancel, Close, etc.

Guess we could hide it on MacOS, but for the settings window the restore defaults button would be pretty lonely down there without a friend.

PR #3477, opened 2025-07-08
Removed analog controller from Tenchu 1

PugsyMAME:


Pesa (commented):
You should keep `AnalogController` and add `DisableAutoAnalogMode` instead (see discord).

PR #3479, opened 2025-07-14
Qt: Refactor `GameListWidget` tool buttons and corresponding menu actions

Pesa:
The main idea is to leverage the power of QAction to keep UI state synchronized in cases where the same action can be performed through multiple UI controls, e.g. a menu item and a toolbar button. Currently DuckStation does most of its UI state management manually.

The only user-visible changes are:
* The "game list" / "game grid" / "system display" menu items are checkable and their state is synchronized with the buttons in GameListWidget. This was the main motivation for this PR. Also, a QActionGroup is used to enforce mutual exclusion.
* The "grid titles" button and "grid scale" slider are hidden when the widget is showing the list view, instead of just being disabled.

I'll do more testing tomorrow but I didn't find anything broken yet.

### DIFF
--- a/src/duckstation-qt/gamelistwidget.ui
+++ b/src/duckstation-qt/gamelistwidget.ui
@@ -54,9 +54,6 @@
        <property name="text">
         <string>Game List</string>
        </property>
-       <property name="icon">
-        <iconset theme="list-check"/>
-       </property>
        <property name="checkable">
         <bool>true</bool>
        </property>
@@ -75,9 +72,6 @@
        </property>
        <property name="text">
         <string>Game Grid</string>
-       </property>
-       <property name="icon">
-        <iconset theme="function-line"/>
        </property>
        <property name="checkable">
         <bool>true</bool>
@@ -98,9 +92,6 @@
        <property name="text">
         <string>Merge Multi-Disc Games</string>
        </property>
-       <property name="icon">
-        <iconset theme="play-list-2-line"/>
-       </property>
        <property name="checkable">
         <bool>true</bool>
        </property>
@@ -119,9 +110,6 @@
        </property>
        <property name="text">
         <string>Show Titles</string>
-       </property>
-       <property name="icon">
-        <iconset theme="price-tag-3-line"/>
        </property>
        <property name="checkable">
         <bool>true</bool>
--- a/src/duckstation-qt/mainwindow.cpp
+++ b/src/duckstation-qt/mainwindow.cpp
@@ -401,7 +401,7 @@ void MainWindow::createDisplayWidget(bool fullscreen, bool render_to_main, bool 
     m_ui.actionViewSystemDisplay->setChecked(true);
   }
 
-  updateDisplayRelatedActions(true, render_to_main, fullscreen);
+  updateDisplayRelatedActions(true, fullscreen);
   updateShortcutActions(false);
 
   updateDisplayWidgetCursor();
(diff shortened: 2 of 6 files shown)


stenzek:
Haven't had a deep look at the changes yet, but from a quick glance make sure you haven't broken the "no games" case (`m_empty_ui`). The games list/grid should not show under any circumstances if it is empty.

Pesa:
> Haven't had a deep look at the changes yet, but from a quick glance make sure you haven't broken the "no games" case (`m_empty_ui`). The games list/grid should not show under any circumstances if it is empty.

I tested all combinations that I could think of. Found a couple of minor bugs but all pre-existing.

I reworked the first commit, the approach is simpler and more robust now. Also fixes a minor glitch where GameListWidget button presses were ignored when the placeholder widget was displayed (i.e. with zero games). Now the buttons and the rest of the toolbar react as expected (widget still isn't switched if there are no games).

The second commit in the series is split from the first, in case you don't like it, it can be dropped independently from the rest. Though I'd like to add another button to toggle game icons in the near future, so I think hiding irrelevant controls would be slightly cleaner/less cluttered.

The third commit fixes a pre-existing bug with the first column of the game table not updating correctly.

Fourth commit was easy enough so why not.

src/duckstation-qt/gamelistwidget.ui
```diff
@@ -54,9 +54,6 @@
        <property name="text">
         <string>Game List</string>
        </property>
-       <property name="icon">
```
Pesa, line 57: the icons for these buttons are inherited from the associated actions now

src/duckstation-qt/mainwindow.cpp
```diff
@@ -428,9 +429,6 @@ void MainWindow::releaseRenderWindow()
 
   updateDisplayRelatedActions(false, false, false);
   updateShortcutActions(false);
-
-  m_ui.actionViewSystemDisplay->setEnabled(false);
-  m_ui.actionFullscreen->setEnabled(false);
```
Pesa: these are already disabled in `updateDisplayRelatedActions` that is called a couple lines above

src/duckstation-qt/mainwindow.cpp
```diff
@@ -296,6 +296,7 @@ std::optional<WindowInfo> MainWindow::acquireRenderWindow(RenderAPI render_api,
     }
 
     updateDisplayWidgetCursor();
+    updateDisplayRelatedActions(!surfaceless, false, fullscreen);
```
Pesa: not 100% sure about the flags here

src/duckstation-qt/mainwindow.cpp
```diff
@@ -297,6 +297,7 @@ std::optional<WindowInfo> MainWindow::acquireRenderWindow(RenderAPI render_api,
     }
 
     updateDisplayWidgetCursor();
+    updateDisplayRelatedActions(!surfaceless, fullscreen);
```
Pesa: this might no longer be needed after b07998512e4222317d54de9343c58f166af80a1f

stenzek: Yep, it should get called through destroy and create.

Pesa: Actually this code path doesn't recreate the display widget, so we still need to call `updateDisplayRelatedActions`. Fixed in #3485.

PR #3484, opened 2025-07-18
Qt: Don't enable window size menu if system isn't running

Pesa:
Starting BPM in a separate window enables the "Window Size" submenu, however the menu items are nonfunctional. This is because the resizing/rescaling functionality is tied to having a running system (see the condition in `EmuThread::requestDisplaySize()` among others). Therefore let's add the additional condition to the UI logic as well.

### DIFF
--- a/src/duckstation-qt/mainwindow.cpp
+++ b/src/duckstation-qt/mainwindow.cpp
@@ -423,7 +423,7 @@ void MainWindow::createDisplayWidget(bool fullscreen, bool render_to_main)
 
 void MainWindow::exitFullscreen(bool wait_for_completion)
 {
-  if (!m_display_widget || !isRenderingFullscreen())
+  if (!isRenderingFullscreen())
     return;
 
   g_emu_thread->setFullscreen(false);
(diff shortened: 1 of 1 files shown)


src/duckstation-qt/mainwindow.cpp
```diff
@@ -423,7 +423,7 @@ void MainWindow::createDisplayWidget(bool fullscreen, bool render_to_main)
 
 void MainWindow::exitFullscreen(bool wait_for_completion)
 {
-  if (!m_display_widget || !isRenderingFullscreen())
+  if (!isRenderingFullscreen())
```
Pesa, line 426: this is just an unrelated cleanup, the check on `m_display_widget` is already inside `isRenderingFullscreen()`

src/duckstation-qt/mainwindow.cpp
```diff
@@ -540,8 +540,7 @@ void MainWindow::onMouseModeRequested(bool relative_mode, bool hide_cursor)
 {
   m_relative_mouse_mode = relative_mode;
   m_hide_mouse_cursor = hide_cursor;
-  if (m_display_widget)
```
Pesa, line 543: similarly here and below, `updateDisplayWidgetCursor()` already checks `m_display_widget`

PR #3487, opened 2025-07-21
Code cleanups

Pesa:


stenzek:
Is it actually worthwhile for e.g. a constructor that takes a span? For pointer or numerical arguments I can understand it, but objects?

Pesa:
The parameter type is not a factor. The general guideline is to declare all constructors that can be called with a single argument (except copy/move constructors) as `explicit` to prevent implicit conversions. Only in those (typically very few) cases where implicit conversions are desired should `explicit` be omitted. Basically this behavior is yet another bad C++ default. That said, it's just a guideline, albeit a widely adopted one. Feel free to drop the commit if you don't like it, or let me know which constructors should be left alone. I'm happy to change.

PR #3493, opened 2025-07-27
Qt: Use localized date/time formatting in a few more places

Pesa:
I acknowledge that the second commit is purely subjective, so feel free to drop it.

### DIFF
--- a/src/core/game_database.h
+++ b/src/core/game_database.h
@@ -107,7 +107,7 @@ struct Entry
   std::string_view publisher;
   std::string_view compatibility_version_tested;
   std::string_view compatibility_comments;
-  u64 release_date;
+  u64 release_date; ///< Number of seconds since Epoch.
   u8 min_players;
   u8 max_players;
   u8 min_blocks;
--- a/src/duckstation-qt/gamesummarywidget.cpp
+++ b/src/duckstation-qt/gamesummarywidget.cpp
@@ -161,8 +161,11 @@ void GameSummaryWidget::populateUi(const std::string& path, const std::string& s
 
     QString release_info;
     if (entry->release_date != 0)
-      release_info =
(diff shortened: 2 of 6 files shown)


src/duckstation-qt/gamesummarywidget.cpp
```diff
@@ -161,8 +161,11 @@ void GameSummaryWidget::populateUi(const std::string& path, const std::string& s
 
     QString release_info;
     if (entry->release_date != 0)
-      release_info =
-        tr("Released %1").arg(QDateTime::fromSecsSinceEpoch(entry->release_date, QTimeZone::utc()).date().toString());
+    {
+      const QString date = QDateTime::fromSecsSinceEpoch(static_cast<qint64>(entry->release_date), QTimeZone::utc())
+                             .toString(QtHost::GetApplicationLocale().dateFormat());
```
Pesa, line 166: Can't use `QtHost::FormatNumber` because it converts to the local timezone before formatting, which is wrong in this case (can result in a different date).

src/core/game_database.h
```diff
@@ -107,7 +107,7 @@ struct Entry
   std::string_view publisher;
   std::string_view compatibility_version_tested;
   std::string_view compatibility_comments;
-  u64 release_date;
+  u64 release_date; ///< Number of seconds since Epoch.
```
Pesa, line 110: IMHO this could be represented with `std::chrono::year_month_day` or `std::chrono::sys_days` (and then converted to `QDate` when needed on the Qt side). Not sure if this is already supported by the compiler/stdlib version combinations you're using though...

src/duckstation-qt/gamesummarywidget.cpp
stenzek: Yeah.. I was thinking of adding local variants to the enum, but since it's only used in this one place, meh.

PR #3495, opened 2025-07-27

zkdpower:
"CD-ROM Disable Speedup on MDEC"

"Disable CD-ROM Speedup on MDEC"

are same? Could you unify it?


stenzek:
Different context. Matches the surrounding text, so not really.

PR #3498, opened 2025-08-07
Qt: Widen the `Size` and `Raw Size` columns in game list view

Pesa:
master:
master

PR:
pr


Pesa:
Yeah, it's almost perfect now. The year column became too narrow (see discord). But this PR is no longer needed.

PR #3507, opened 2025-08-12
Qt: Fix crash toggling achievements in per-game settings

Pesa:
`loginBox` and everything inside it is deleted in the per-game version of `AchievementSettingsWidget`.

### DIFF
--- a/src/duckstation-qt/achievementsettingswidget.ui
+++ b/src/duckstation-qt/achievementsettingswidget.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>674</width>
-    <height>479</height>
+    <height>524</height>
    </rect>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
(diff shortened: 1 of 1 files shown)


src/duckstation-qt/achievementsettingswidget.cpp
```diff
@@ -82,7 +82,6 @@ AchievementSettingsWidget::AchievementSettingsWidget(SettingsWindow* dialog, QWi
     tr("Shows a popup in the lower-right corner of the screen when progress towards a measured achievement changes."));
 
   connect(m_ui.enable, &QCheckBox::checkStateChanged, this, &AchievementSettingsWidget::updateEnableState);
-  connect(m_ui.hardcoreMode, &QCheckBox::checkStateChanged, this, &AchievementSettingsWidget::updateEnableState);
```
Pesa, line 85: this seems unnecessary, nothing in `updateEnableState()` depends on the hardcore mode checkbox.

stenzek: It might have been a leftover from my old cheevos implementation where leaderboards were toggleable.

PR #3524, opened 2025-08-19
GameDB: Set max speedup cycles to 70k for `Resident Evil 3`

Pesa:
@6lackmag3


data/resources/gamedb.yaml
```diff
@@ -15896,6 +15896,8 @@ SLPS-02300:
     - DisableWidescreen
     - ForcePGXPCPUMode # Fixes jitter in character models.
     - ForceRecompilerICache # Fixes hangs in CD code.
```
Pesa, line 15898: I wonder if this is still necessary. I played maybe 1/3 of the game without ICache and with 70k cycles (and max speedup), didn't experience any issues.

PR #3531, opened 2025-08-26
Enable `PGXPPreserveProjFP` for `Ghost in the Shell`

Pesa:
| PGXPPreserveProjFP disabled | PGXPPreserveProjFP enabled |
| :---: | :---: |
| off-A | on-A |
| off-B | on-B |


data/resources/gamedb.yaml
```diff
@@ -66876,9 +66885,11 @@ SLUS-00552:
     multitap: false
     linkCable: false
 SCPS-45092:
-  name: "Ghost in the shell"
+  name: "Ghost in the Shell (Asia)"
```
Pesa, line 66888: Didn't find much info online about this version, but it's listed under "Undumped Asia Releases" on the [redump wiki](http://wiki.redump.org/index.php?title=Sony_PlayStation_-_Asia_Undumped_Discs) and also appears [here](https://gamefaqs.gamespot.com/ps/197447-ghost-in-the-shell/data) and [here](https://www.retroplace.com/en/games/57068--ghost-in-the-shell) as "Asia".

PR #3533, opened 2025-08-26

zkdpower:


stenzek:
It's not only for developers, there might be some SBCs where it is preferred.

PR #3542, opened 2025-08-31
fix off-by-1 error for memcard editor icon size

ANK-dev:


ANK-dev:
Fixed by 44908fe and 126171f.

PR #3547, opened 2025-09-01

zkdpower:
report a bug of 0.1-9611 version:
Taking Resident Evil 2 (as shown in the image) as an example, when loading a save file from the game list in "Big Picture Mode", there is a certain probability that glitches will appear like the one illustrated here.


zkdpower:
the bug still occurs in 0.1-9619 version

PR #3561, opened 2025-09-09
Qt: Disable UI of settings forced via gamedb, part two

Pesa:
Sorry, missed these in #3558.

`gpu_pgxp_preserve_proj_fp` is treated specially because it can be set to true or false in the db.

### DIFF
--- a/src/duckstation-qt/graphicssettingswidget.cpp
+++ b/src/duckstation-qt/graphicssettingswidget.cpp
@@ -127,6 +127,8 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsWindow* dialog, QWidget* 
                                        !m_dialog->hasGameTrait(GameDatabase::Trait::DisableTextureFiltering) ||
                                          !m_dialog->hasGameTrait(GameDatabase::Trait::DisableSpriteTextureFiltering));
   SettingWidgetBinder::SetAvailability(m_ui.pgxpEnable, !m_dialog->hasGameTrait(GameDatabase::Trait::DisablePGXP));
+  SettingWidgetBinder::SetAvailability(m_ui.pgxpDepthBuffer,
+                                       !m_dialog->hasGameTrait(GameDatabase::Trait::DisablePGXPDepthBuffer));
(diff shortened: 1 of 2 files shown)


src/duckstation-qt/graphicssettingswidget.cpp
```diff
@@ -205,14 +207,18 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsWindow* dialog, QWidget*
...
                                        !m_dialog->hasGameTrait(GameDatabase::Trait::DisablePGXPCulling));
-  SettingWidgetBinder::SetAvailability(m_ui.pgxpPreserveProjPrecision,
-                                       !m_dialog->hasDatabaseEntry() ||
-                                         m_dialog->getDatabaseEntry()->gpu_pgxp_preserve_proj_fp.value_or(true));
   SettingWidgetBinder::SetForceEnabled(m_ui.pgxpCPU, m_dialog->hasGameTrait(GameDatabase::Trait::ForcePGXPCPUMode));
   SettingWidgetBinder::SetForceEnabled(m_ui.pgxpVertexCache,
                                        m_dialog->hasGameTrait(GameDatabase::Trait::ForcePGXPVertexCache));
   SettingWidgetBinder::SetForceEnabled(m_ui.pgxpDisableOn2DPolygons,
                                        m_dialog->hasGameTrait(GameDatabase::Trait::DisablePGXPOn2DPolygons));
+  if (auto dbentry = m_dialog->getDatabaseEntry(); dbentry && dbentry->gpu_pgxp_preserve_proj_fp.has_value())
```
stenzek, line 215: Seems like this should be in a helper functions, since there will be other traits that behave the same way such as those above it.

Pesa: isn't `gpu_pgxp_preserve_proj_fp` the only one with an explicit true/false value in the db?

stenzek: I was more thinking from the perspective of "results in a forced value", e.g. force software renderer. But eh, I guess the availability kinda handles it.

PR #3563, opened 2025-09-10
Qt: Show tooltip when adjusting icon/cover scale

ANK-dev:
Display a tooltip with the current size when adjusting the scale slider for icons or covers.

https://github.com/user-attachments/assets/ae15548b-82ad-42d2-a512-8bf1fdc02400

### DIFF
--- a/src/duckstation-qt/gamelistwidget.cpp
+++ b/src/duckstation-qt/gamelistwidget.cpp
@@ -1373,13 +1373,8 @@ void GameListWidget::initialize(QAction* actionGameList, QAction* actionGameGrid
   m_ui.showGridTitles->setDefaultAction(actionGridShowTitles);
   m_ui.showLocalizedTitles->setDefaultAction(actionShowLocalizedTitles);
 
-  // Configure scale slider for hover events
-  m_ui.scale->setAttribute(Qt::WA_Hover, true);
-  m_ui.scale->installEventFilter(this);
-
   connect(m_ui.scale, &QSlider::sliderPressed, this, &GameListWidget::showScaleToolTip);
   connect(m_ui.scale, &QSlider::valueChanged, this, &GameListWidget::onScaleSliderChanged);
-
   connect(m_ui.filterType, &QComboBox::currentIndexChanged, this, [this](int index) {
     m_sort_model->setFilterType((index == 0) ? GameList::EntryType::MaxCount :
                                                static_cast<GameList::EntryType>(index - 1));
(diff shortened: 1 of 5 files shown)


src/duckstation-qt/gamelistwidget.cpp
```diff
@@ -1371,7 +1373,13 @@ void GameListWidget::initialize(QAction* actionGameList, QAction* actionGameGrid
   m_ui.showGridTitles->setDefaultAction(actionGridShowTitles);
   m_ui.showLocalizedTitles->setDefaultAction(actionShowLocalizedTitles);
 
+  // Configure scale slider for hover events
+  m_ui.scale->setAttribute(Qt::WA_Hover, true);
```
Pesa: is this necessary? can you use QEnterEvent instead of hover?

src/duckstation-qt/gamelistwidget.cpp
```diff
@@ -1718,6 +1726,36 @@ void GameListWidget::setViewMode(int stack_index)
...
+      return false;
+  }
+}
+
+void GameListWidget::showScaleToolTip()
+{
+  const int value = m_ui.scale->value();
+  if (isShowingGameGrid())
+    QToolTip::showText(QCursor::pos(),
+                       QString("Cover scale: %1%").arg(static_cast<float>(value) / DEFAULT_COVER_SCALE, 0, 'f', 0));
```
Pesa: why divide by `DEFAULT_COVER_SCALE`? for covers, I think the value is already the scale from 10% to 200%.

Also, the string needs to be translatable, see `tr()`.

ANK-dev: The default scale value from a fresh install is 45 so by dividing it by 0.45 I display it as 100%.

Pesa: but why is 0.45 treated as 100%? shouldn't 1.0 be 100%?
with the division you get a different value in the UI vs the config file, and the max is a weird number like 444% 😛

src/duckstation-qt/gamelistwidget.cpp
ANK-dev: The hover event was removed.

src/duckstation-qt/gamelistwidget.cpp
ANK-dev: The scale percentage display value was adjusted to match the config file (10%-200%).

PR #3574, opened 2025-09-18
Use DualSense Mic Mute LED for Analog Mode

ANK-dev:
Provide an option to use the Mic Mute LED present in the DualSense and DualSense Edge controllers as an Analog Mode indicator similar to the Analog Mode LED on the DualAnalog/DualShock controllers.

image

|image|image|
|-|-|

### DIFF
--- a/src/core/analog_controller.cpp
+++ b/src/core/analog_controller.cpp
@@ -29,7 +29,10 @@ AnalogController::AnalogController(u32 index) : Controller(index)
   m_rumble_config.fill(0xFF);
 }
 
-AnalogController::~AnalogController() = default;
+AnalogController::~AnalogController()
+{
+  InputManager::SetGamepadAnalogLED(m_index, false);
+}
 
 ControllerType AnalogController::GetType() const
 {
@@ -334,6 +337,8 @@ void AnalogController::SetAnalogMode(bool enabled, bool show_message)
     return;
 
   m_analog_mode = enabled;
+
+  InputManager::SetGamepadAnalogLED(m_index, enabled);
 
   INFO_LOG("Controller {} switched to {} mode.", m_index + 1u, m_analog_mode ? "analog" : "digital");
   if (show_message)
--- a/src/util/input_manager.cpp
+++ b/src/util/input_manager.cpp
@@ -5,6 +5,10 @@
 #include "imgui_manager.h"
 #include "input_source.h"
 
+#ifdef ENABLE_SDL
+#include "sdl_input_source.h"
+#endif
+
 #include "core/controller.h"
 #include "core/host.h"
 #include "core/system.h"
@@ -1542,6 +1546,7 @@ void InputManager::SetDefaultSourceConfig(SettingsInterface& si)
   si.SetBoolValue("InputSources", "SDL", true);
   si.SetBoolValue("InputSources", "SDLControllerEnhancedMode", false);
   si.SetBoolValue("InputSources", "SDLPS5PlayerLED", false);
+  si.SetBoolValue("InputSources", "SDLPS5MicMuteLEDForAnalogMode", false);
   si.SetBoolValue("InputSources", "XInput", false);
   si.SetBoolValue("InputSources", "RawInput", false);
 }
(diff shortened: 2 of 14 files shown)


src/util/sdl_input_source.cpp
```diff
@@ -337,6 +340,60 @@ u32 SDLInputSource::ParseRGBForPlayerId(std::string_view str, u32 player_id)
   return color;
 }
 
+bool SDLInputSource::isPS5Controller(SDL_Gamepad* gp)
```
Pesa: The surrounding code uses PascalCase, you may want to rename this function to conform.

src/util/sdl_input_source.cpp
```diff
@@ -337,6 +340,60 @@ u32 SDLInputSource::ParseRGBForPlayerId(std::string_view str, u32 player_id)
...
+{
+  const char* supported_controllers[] = {"DualSense Wireless Controller", "DualSense Edge Wireless Controller"};
+
+  bool supported = false;
+  for (auto& supported_controller : supported_controllers)
+  {
+    supported |= (strcmp(supported_controller, SDL_GetGamepadName(gp)) == 0);
```
Pesa: move `SDL_GetGamepadName` outside the loop

src/util/sdl_input_source.cpp
ANK-dev: Fixed

src/util/sdl_input_source.cpp
ANK-dev: Fixed

src/util/input_manager.cpp
```diff
@@ -1784,6 +1791,71 @@ std::unique_ptr<ForceFeedbackDevice> InputManager::CreateForceFeedbackDevice(con
   return {};
 }
 
+// ------------------------------------------------------------------------
+// Analog LED
+// ------------------------------------------------------------------------
+
+void InputManager::SetGamepadAnalogLED(u32 player_index, bool enabled)
+{
+#ifdef ENABLE_SDL
```
stenzek: `InputManager` should not have any API-specific logic in it - that's why the `InputSource` abstraction exists.

You also don't want to be constantly querying the config, since that involves a mutex lock.

Personally, I'm of the opinion that this should be abstracted by SDL, not the application. We could patch SDL, or better yet upstream it.

src/core/analog_controller.cpp
```diff
@@ -29,7 +29,10 @@ AnalogController::AnalogController(u32 index) : Controller(index)
   m_rumble_config.fill(0xFF);
 }
 
-AnalogController::~AnalogController() = default;
+AnalogController::~AnalogController()
+{
+  InputManager::SetGamepadAnalogLED(m_index, false);
```
stenzek: This is the index of the emulated controller, which does not necessarily correspond to the player index of the connected controller. Emulated controller 0 can be mapped to SDL player 1, or vice versa, etc. So really this should have a bind and get forwarded through, similar to vibration.

src/util/sdl_input_source.cpp
```diff
@@ -337,6 +340,61 @@ u32 SDLInputSource::ParseRGBForPlayerId(std::string_view str, u32 player_id)
   return color;
 }
 
+bool SDLInputSource::IsPS5Controller(SDL_Gamepad* gp)
+{
+  const char* supported_controllers[] = {"DualSense Wireless Controller", "DualSense Edge Wireless Controller"};
```
stenzek: I feel like there has to be a better way to identify this. Seems error-prone and/or not future proof - I know my PS5 controller has appeared under different device name depending on SDL hints.

ANK-dev: I've changed it to use the Vendor & Product ID.

src/util/input_manager.cpp
ANK-dev: I've removed the API-specific logic and changed it to binds.

src/core/analog_controller.cpp
ANK-dev: The method now uses binds to control the LED on the correct physical controller.

src/util/input_manager.cpp
ANK-dev: Regarding patching SDL or upstreaming, I'm open to it, however I'm not familiar with the process.

PR #3579, opened 2025-09-28

zkdpower:
It seems that the "animate game icons" aren't showing their effects. How to use it?


stenzek:
You need to delete the "old" icons in `gameicons` for it to save new APNG icons.

PR #3699, opened 2026-02-26
[CI] : Add Mini builds for x64 and x64 SSE2

rtissera:
This PR adds support for building DuckStation-Mini to x86_64 and x86_64 SSE2 targets.

### DIFF
--- a/.github/workflows/linux-appimage-build.yml
+++ b/.github/workflows/linux-appimage-build.yml
@@ -19,10 +19,17 @@ jobs:
             asset: "DuckStation-x64.AppImage"
             artifact: "linux-x64-appimage"
             cmakeoptions: ""
+            packagescript: "scripts/appimage/make-appimage.sh"
           - name: "x64 SSE2"
             asset: "DuckStation-x64-SSE2.AppImage"
             artifact: "linux-x64-sse2-appimage"
             cmakeoptions: "-DDISABLE_SSE4=ON"
+            packagescript: "scripts/appimage/make-appimage.sh"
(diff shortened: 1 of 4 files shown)


stenzek:
I'm a little confused about the use case here - mini is only intended for systems that aren't running a window manager, and is missing a ton of functionality (most important being the ability to update itself). Like super low spec arm boards for example, and afaik there aren't any x86 machines like this.

Not 100% against the idea, but it would need to have a use case that's not already covered.

rtissera:
I understand your valid point.
I felt like the mini flavor could fit well on Steam-Deck-like handhelds or constrained N100-class devices.

stenzek:
Don't those still have a window manager/compositor? I don't own any of these, but I was under the impression that SteamOS etc ran gamescope even in game mode.

stenzek:
Performance-wise it's exactly the same code running. Startup time is actually slightly worse in the mini variant since it needs to create the GPU device for BPM. The Widgets UI doesn't need to.

Basically, running the Qt appimage with `-bigpicture` is no different to Mini, unless you want to run without a WM.

There's also some other issues, like fullscreen mode switching not working because SDL doesn't let you disconnect a GL context from a window, that work fine in the full build.

stenzek:
Think we can safely close this now. I dropped mini almost 3 months ago and haven't heard from anyone being inconvenienced by it, which supports the argument that it was redundant in the first place. Missing so many features was a pretty big dealbreaker, and it would be silly to reimplement them all twice.

PR #3713, opened 2026-04-15
CI: Bump Xcode to 26.4

Pesa:


stenzek:
This will need to wait until after Qt 6.11.1 and the deps are updated, as there's an issue with one of the arm64 intrinsic names that prevents building with 26.4.

PR #3714, opened 2026-04-18
Qt: Two trivial fixes

Pesa:


src/duckstation-qt/advancedsettingswidget.cpp
```diff
@@ -128,7 +128,7 @@ void AdvancedSettingsWidget::onClearWebCacheClicked()
   Error error;
   if (!HTTPCache::Clear(&error))
   {
-    QtUtils::AsyncMessageBox(this, QMessageBox::Critical, "Failed to clear web cache"_L1,
+    QtUtils::AsyncMessageBox(this, QMessageBox::Critical, tr("Failed to clear web cache"),
```
stenzek, line 131: This was deliberate - no point translating error messages which are basically never going to be seen (only with a failing disk), and the content of the error won't be translated anyway.

Pesa: Well, ok, seems odd to have this msgbox treated specially (all others are translated), but I don't really care :)

Feel free to cherry pick the other commit if it's ok.

stenzek: Yea, fair point.. I probably shouldn't have translated the others in the beginning.

stenzek:
Cherry-picked in https://github.com/stenzek/duckstation/commit/7bcafb44f5b60866fec51cd3b8f4cd47eb81d00e

PR #3717, opened 2026-04-20
MMPXEnhanced v3.0

crashGG:
This was a major update nine months later.
Over the course of nine months, this shader has undergone numerous updates and iterations, and is now at version 3.0.

lots of changes:

Use approximate pixels to capture more details
Use mixed to generate intermediate color pixels to create more texture transition details
Improved shape judgment, basically eliminating the bubble defects of the original version.
Almost rewrote most of the code.
Expanded morphological functions allow for greater control over details. The refactoring of the core algorithm (almost) completely eliminates bubbles generated by 4-pixel intersections.

Prior to this PR, the shader had undergone extensive testing on other platforms.

The libretro port of this shader is available at [https://github.com/libretro/slang-shaders/blob/master/edge-smoothing/scalenx/shaders/mmpx-ex.slang] welcome feedback from retro game enthusiasts.

The ppsspp port of this shader is available at https://github.com/hrydgard/ppsspp/blob/master/assets/shaders/tex_mmpx_adv.csh

It's worth noting that due to the addition of a large amount of conditional logic, this shader has become quite heavy. It's fine for 2x upscale in  2D games, but for  4x or higher in 3D games, a reasonably faster GPU is required.

### DIFF
--- a/src/core/gpu_hw_shadergen.cpp
+++ b/src/core/gpu_hw_shadergen.cpp
@@ -835,167 +835,553 @@ void FilteredSampleFromVRAM(TEXPAGE_VALUE texpage, float2 coords, float4 uv_limi
   }
   else if (texture_filter == GPUTextureFilter::MMPXEnhanced)
   {
-    ss << "#define src(xoffs, yoffs) packUnorm4x8(SampleFromVRAM(texpage, bcoords + float2((xoffs), (yoffs)), "
-          "uv_limits))\n";
-
-    /*
-     * This part of the shader is from MMPX.glc from https://casual-effects.com/research/McGuire2021PixelArt/index.html
-     * Copyright 2020 Morgan McGuire & Mara Gagiu.
-     * Provided under the Open Source MIT license https://opensource.org/licenses/MIT
-     */
     ss << R"(
-uint luma(uint C) {
-    uint alpha = (C & 0xFF000000u) >> 24;
-    return (((C & 0x00FF0000u) >> 16) + ((C & 0x0000FF00u) >> 8) + (C & 0x000000FFu) + 1u) * (256u - alpha);
+	#define srcf(xoffs,yoffs) SampleFromVRAM(texpage, bcoords + float2((xoffs), (yoffs)), uv_limits)
+	#define src(xoffs,yoffs) packUnorm4x8(srcf(xoffs,yoffs))
+	)";
+
+	/* MMPXEnhanced v3.0
+	 * This shader is an enhanced iteration of MMPX.glc (see above).
+	 * It improves the visual quality while preserving the pixel-art aesthetic by 
+	 * identifying and analyzing specific geometric shapes, effectively resolving 
+	 * the artifacts found in the original algorithm.
+	 * 
+	 * © 2025-2026 by crashGG.
+	 * Licensed under the same terms as MMPX.glc.
+	 */
+
+    ss << R"(
+
+//RGB visual weight + alpha segmentation
+float luma(float4 col) {
+
+	//Use CRT-era BT.601 standard. Clamp range to [0.0 - 0.999]
+    float rgbsum =min(dot(col.rgb, float3(0.299, 0.587, 0.114)), 0.9999);
+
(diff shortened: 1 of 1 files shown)


stenzek:
Nice work! Do you think it's worth keeping the old version around, given the differences in GPU load?

crashGG:
> Nice work! Do you think it's worth keeping the old version around, given the differences in GPU load?

now, mmpx enhance solution is already mature. There's no need to keep the initial modified version from nine months ago. If GPU overhead is a concern, user can use mmpx. Another point that bothers me is the relatively long compilation time of this shader on DuckStation. For my hardware 7840hs + 780m , the compilation time for Libretro Vulkan is about 2 seconds, OpenGL about 3 seconds, for PPSSPP about 3 seconds. but on duckstation about 20~40 seconds(vulkan,dx11,opengl)

stenzek:
But is there any advantage to the current modified version (v2)? For users who can't run v3, but want an improvement over v1?

The compilation times are expected to be higher. It's not compiling the shader once, but tens of times for every possible draw configuration. That way it avoids shader compilation stutter at runtime. It's cached, so it only needs to do it once. I guess I could parallelise it, but never really bothered since it seemed fast enough.

The DX11 compiler is _much_ slower than the DX12/GL(most drivers)/Vulkan compilers (it's much older). But code quality is similar.

crashGG:
> But is there any advantage to the current modified version (v2)? For users who can't run v3, but want an improvement over v1?
> 
In fact, starting with version 1, each iteration has been accompanied by algorithm modifications and feature additions. This has ultimately led to the heavy today.I think I can make a lightweight, streamlined version based on this, removing approximate pixel capture and only fixing graphics bugs, with the final GPU overhead being roughly the same as the original version.

stenzek:
I'm happy with whatever you think is best. Generally speaking, users tend to get cranky when something is "replaced" with something that has a greater performance cost, even if it is objectively better in all ways.

I'd suggest something like:
- MMPX: original version
- MMPX (Enhanced): current version in master
- MMPX (Enhanced 3.0): your PR here

Or perhaps "ultimate" or some other word.

If you don't want to change the other files I'm happy to do it.

crashGG:
This enhanced 3.0 is now complete.I’m currently working on a high-performance hybrid version that fuses the core logic of the original MMPX with the Enhanced 3.0 implementation. It resolves the original’s visual artifacts while delivering blazing-fast throughput — in fact, it’s faster than the original, or at the very least no slower.

So going forward, we would have three versions coexisting:
1.Original MMPX
2.The bug-fixed, high-speed MMPX
3.The higher-cost, full-quality MMPX Enhanced

For version 2, I was thinking of names like MMPX Remaster or MMPX Alternative.
For version 3, maybe MMPX Advanced, Enhanced, or Ultimate?

English isn’t my native language, so I`d love your take on which naming sounds cleanest and most technically appropriate 。
 you can merge it first. Then change it to the name you deem most suitable.

kamfretoz:
1. MMPX
2. MMPX Fast
3. MMPX Full (Slower)

Just my 2 cents

stenzek:
Based on the points you listed:

1. MMPX (Original)
2. MMPX (Performance)
3. MMPX (Quality)

Would make sense to me. Totally adapted from upscaling tech, but it's terms users are familiar with.

crashGG:
> Based on the points you listed:
> 
> 1. MMPX (Original)
> 2. MMPX (Performance)
> 3. MMPX (Quality)
> 
> Would make sense to me. Totally adapted from upscaling tech, but it's terms users are familiar with.

ok.
Looking forward to your improvements to shader parallel compilation.

stenzek:
I've left the existing version as "Enhanced" and named this new version "Quality" for now. Can change enhanced->performance later, or keep it and add the new performance version.

stenzek:
FWIW, shader compilation is around 5-10 seconds with DX12/Vulkan on my 7950X3D/RTX5080. I'll still parallelise it since it's quite slow on MacOS in general (the Metal compiler is slower than everything else), but will need to wait until I have a larger block of time free from work to do that.

PR #3720, opened 2026-04-22
SPU: Display friendly reverb mode names in Debug State window

N4gtan:
This PR adds a helper function to identify and display the current reverb mode in the Debug SPU State window.

It checks both `FB_SRC_A` and `IIR_COEF` registers to ensure the preset matches. If the registers don't match a known preset, it falls back to "Unknown".

### DIFF
--- a/src/core/spu.cpp
+++ b/src/core/spu.cpp
@@ -2377,6 +2377,27 @@ void SPU::ProcessReverb(s32 left_in, s32 right_in, s32* left_out, s32* right_out
 #endif
 }
 
+static constexpr const char* GetReverbModeName(const SPU::ReverbRegisters& reverb_registers)
+{
+  const u32 signature = (static_cast<u32>(reverb_registers.FB_SRC_A) << 16) |
+                        (static_cast<u16>(reverb_registers.IIR_COEF));
+
+  switch (signature)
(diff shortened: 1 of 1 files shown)


stenzek:
Please don't reference psyq here. DuckStation contains no SDK code or referencces, and I'm not going to change that for legal reasons.

You should use your own descriptions.

N4gtan:
Ok, np I will delete that

stenzek:
Not a question of deleting, it needs to be named based on your own descriptions.

If it was something more significant I'd just close it entirely since I don't want the legal risk, but AFAIK numbers can't really be copyrighted.

I might do that anyway, will have to give it some thought.

N4gtan:
Understood. I leaved the names with generic acoustic descriptions (e.g., 'Small Studio', 'Hall', 'Space Echo'). These are standard audio industry terms and are not tied to any specific SDK or copyrighted material. I've also removed all comments and references to the SDK headers to ensure the codebase remains independent.

stenzek:
Cherry-picked in https://github.com/stenzek/duckstation/commit/70ec9c4915a02aec94db455fdb3f6941fc511aa3 due to conflicts.

PR #3729, opened 2026-04-28
GameDB: ForceInterpreter for Battle Arena Toshinden

6lackmag3:


stenzek:
Sorry, this isn't something I'm comfortable with - overclock breaks things regardless, int should only be forced when the game is doing something silly that breaks the dynarec (e.g. stale icache execution shenanigans). You can of course set this yourself in game properties.

6lackmag3:
Oh, thanks for explanation. I'll keep that in mind.

PR #3731, opened 2026-05-01
Modern Lossless Dynamic Quantization Scanline Shader

crashGG:
This shader is a port of my scanlines-Modern.slang libretro version.
https://github.com/libretro/slang-shaders/blob/master/scanlines/shaders/scanlines-Modern.slang

This shader is not designed to simply simulate the scanline + cross grid effect of old CRT monitors. 
Instead, it aims to combine the advantages of sharp clarity on modern displays with retro games, 
enabling better pixel-level scaling.
The generation intensity of scanlines is dynamically quantized and adjusted based on the human eye's 
perceptual curve for chromatic brightness, rather than using rigid stripe overlay.

Core Features:
- Supports independent adjustment of vertical scanline and horizontal crossline intensity/density, 
  adapting to different resolutions (1080P/4K / high-magnification scaling);
- Default parameters are suitable for most pixel games scaled up on large modern 4K resolution screens, 
  with lossless brightness/color;
- Optimized scanline performance based on human eye brightness sensitivity curve: 
  scanlines are prominent in medium brightness areas and weakened in extreme brightness areas;
- Adjustable color channel quantization attenuation, suitable for games like GBA that have 
  hardware-encoded gamma bias (over-bright/grayed out) due to lack of backlighting. 
  It can perfectly restore vivid colors. 

    Perceptual Sensitivity Curve:
    Sensitivity
      ↑
      |        Peak Sensitivity (0.4-0.6 Luma)
      |           / \
      |          /   \
      |         /     \____ High-luma roll-off
      |        /
      +--------------------→ Luminance (0.0 - 1.0)

### DIFF
--- a/data/resources/shaders/reshade/Shaders/scanline-modern-ex.fx
+++ b/data/resources/shaders/reshade/Shaders/scanline-modern-ex.fx
@@ -1,0 +1,139 @@
+#include "ReShade.fxh"
+
+/*
+
+This shader is not designed to simply simulate the scanline + cross grid effect of old CRT monitors. 
+Instead, it aims to combine the advantages of sharp clarity on modern displays with retro games, 
+enabling better pixel-level scaling.
+The generation intensity of scanlines is dynamically quantized and adjusted based on the human eye's 
+perceptual curve for chromatic brightness, rather than using rigid stripe overlay.
+
+Core Features:
+- Supports independent adjustment of vertical scanline and horizontal crossline intensity/density, 
+  adapting to different resolutions (1080P/4K / high-magnification scaling);
+- Default parameters are suitable for most pixel games scaled up on large modern 4K resolution screens, 
+  with lossless brightness/color;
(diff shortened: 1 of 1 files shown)


crashGG:
When paired with PS1-style low-poly assets, this shader renders an exquisite, granular texture on my 27-inch 4K display. It strikes a sophisticated balance between retro-aliasing aesthetics and modern high-fidelity clarity, creating a visual depth that is both nostalgic and refined. so I recommended it.

MediEvil 2026-05-01-19-00-26

crashGG:
Tekken 2 2026-05-01-19-28-51

stenzek:
FWIW, you should also be able to load the slang version as-is. If it doesn't work, then that's a bug.

crashGG:
> FWIW, you should also be able to load the slang version as-is. If it doesn't work, then that's a bug.

I completed this shader code in November 2025. Back then, DuckStation didn't seem to support Slang.

PR #3742, opened 2026-05-17
optimization scanline-modern shader

crashGG:
optimization scanline-modern shader


crashGG:
Improvement: The interval frequency of scan lines and cross lines is now set based on the number of physical pixels on the screen, making it more intuitive.
 One bug  fixed.

PR #3747, opened 2026-06-01
Qt: Fix overlapping widgets in game summary

Pesa:
I think we can remove `minimumSize` completely since QLineEdit uses the Expanding size policy by default and the parent layout will do the right thing.

Avoids the issue shown below:
overlap


stenzek:
Nice catch.

PR #3749, opened 2026-06-05
GameDB: Add metadata to `King's Field 3 - Pilot Style`

Pesa:
Reported by lukkasz323 on discord. Also fixed date/genre of a few other entries.


data/resources/gamedb.yaml
```diff
@@ -90082,40 +90082,39 @@ SLPS-00247:
     vibration: false
     multitap: false
     linkCable: false
-SCES-00510:
+SLPS-00017:
```
Pesa, line 90085: The first part of the diff is unreadable because I moved the two KF1 entries before KF2. The only actual change in the moved entries is Shooter -> Action in `genre`.

PR #3757, opened 2026-06-22
Spanish (Spain) translation update 2026/06/22

IlDucci:
- Undoing the previous mistake.
- Translating new strings.
- Abbreviated the Achievements hint on QT Mode to make sure it fits on the box and thus, the hint that says "Press Shift + Mouse Wheel to see longer hints" is seen from the get-go.
- Fixing some consistency issues with "Deshabilitar"/"Desactivar".

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_es-ES.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_es-ES.ts
@@ -170,79 +170,79 @@ No obstante, el modo «hardcore» también impide el uso de los guardados rápid
 <context>
     <name>AchievementSettingsWidget</name>
     <message>
-        <location filename="../achievementsettingswidget.ui" line="90"/>
+        <location filename="../achievementsettingswidget.ui" line="98"/>
         <location filename="../achievementsettingswidget.cpp" line="75"/>
         <source>Enable Achievements</source>
         <translation>Habilitar logros</translation>
     </message>
     <message>
-        <location filename="../achievementsettingswidget.ui" line="97"/>
+        <location filename="../achievementsettingswidget.ui" line="105"/>
         <location filename="../achievementsettingswidget.cpp" line="77"/>
         <source>Enable Hardcore Mode</source>
         <translation>Habilitar modo «hardcore»</translation>
     </message>
     <message>
-        <location filename="../achievementsettingswidget.ui" line="118"/>
+        <location filename="../achievementsettingswidget.ui" line="126"/>
         <location filename="../achievementsettingswidget.cpp" line="86"/>
         <source>Test Unofficial Achievements</source>
         <translation>Probar logros no oficiales</translation>
     </message>
     <message>
-        <location filename="../achievementsettingswidget.ui" line="245"/>
+        <location filename="../achievementsettingswidget.ui" line="253"/>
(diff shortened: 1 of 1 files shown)


stenzek:
Need to rebase on master, or I can manually push it later

IlDucci:
Could you please push it later? I'll probably break something during the rebase (I'm that of a Git noob).

fox140cv:
You're the only one who translates the increase in scales of "x" as "×". In all other languages ​​it's always "x". Example: "Nativa 2x"

And visually it looks better with the x

These screenshots show how the translation will look in the current PR:
cap2
cap23
__________
These screenshots show what it currently looks like:

cap1
cap-1

There is clearly a regression if it merges.

IlDucci:
> You're the only one who translates the increase in scales of "x" as "×". In all other languages ​​it's always "x". Example: "Nativa 2x"
> 
> And visually it looks better with the x
> 
> These screenshots show how the translation will look in the current PR: 
> 
> These screenshots show what it currently looks like:
> 
> There is clearly a regression if it merges.

I cannot and should not give an opinion regarding other languages because I do not know their actual rules in regards to multiplication signs, nor I do not know if their translators actually know what are they doing in regards to that. I can only focus on Spanish.

Using a lowercase x as a multiplication sign is a valid alternative, but the proper thing is to use the actual multiplication sign (Alt + 0215).

References: 
https://en.wikipedia.org/wiki/Multiplication_sign
https://es.wikipedia.org/wiki/Signo_de_multiplicaci%C3%B3n
Windows 10 resolution selector:
imagen

Both the quotation signs and the multiplication replacements made by fox140cv are purely preferential changes or based on what's literally written in English, without considerations regarding the writing of crude anglicisms in Spanish.

Changing "Sincronizar con frec. act. del equipo" por "Sincronizar con frecuencia de pantalla", however, is not a bad way to avoid excessive abbreviation on that particular string, though I think losing "host" ("equipo" in this particular case) is a problem.

fox140cv:
I understand what you mean about the Anglicisms; seeing all the scales with "x" makes it more familiar.

Aside from these minor differences, all instances of "native x" still need to be translated. Otherwise, the dropdown menu looks odd.
And I would reconsider avoiding abbreviations since the layman would not understand that description well.

IlDucci:
The loss of all the remaining "Xx Native" strings is something that I do not know how to fix. I followed stenzek's create-update-and-edit-language.bat and those strings were removed from the translation file.

fox140cv:
You have it in my pull. Add this on line 13184:

cap1

If you can't do it, I'll do it no problem.

IlDucci:
> You have it in my pull. Add this on line 13184:

Yes, that would be the "hackish" way, but if the official bat file removes these strings from the TS file, this issue will reappear at any other language, whenever their translators use the bat file. Something needs to be done elsewhere.

stenzek:
If the original string does not use non-alphsnumeric symbols, then the translation should not either. That actually annoys me because it adds inconsistency, pt-br is the same. MIght do a search/replace pass.

stenzek:
Committed in https://github.com/stenzek/duckstation/commit/553b133b9ce8379ebe1359f3f08a6dbdaeb4d350

PR #3765, opened 2026-07-04
Qt: Respect `SH_Widget_Animation_Duration` + Add missing singular forms

Pesa:


src/duckstation-qt/animatedstackedwidget.cpp
```diff
@@ -87,7 +87,7 @@ void AnimatedStackedWidget::setCurrentIndex(int index)
 
   finishAnimation();
 
-  if (!isVisible() || !style()->styleHint(QStyle::SH_Widget_Animate, nullptr, this))
```
Pesa, line 90: `SH_Widget_Animate` is deprecated.

PR #3767, opened 2026-07-05
Qt: Cancel search when pressing Esc

Pesa:
Add a new `SearchBox` component to encapsulate the common logic and reduce duplication.


stenzek:
Seems a bit overkill for overriding one method... so many regrets of not using QML.

PR #3770, opened 2026-07-05
Update translations: ES, ja, pt-BT, sv and zh-CN

fox140cv:
The 4385 words of translation have been 100% completed.: ES, ja, pt-BT, sv and zh-CN

Osolete translations have been removed. They have been verified to be working correctly.


stenzek:
I'm a little confused at the purpose of this PR. Please don't run lupdate against other translations - it just creates noise and conflicts for the maintainers of those files.

PR #3786, opened 2026-08-21
Update duckstation-qt_es.ts

PeterDelta:
Spanish-international language update

### DIFF
--- a/src/duckstation-qt/translations/duckstation-qt_es.ts
+++ b/src/duckstation-qt/translations/duckstation-qt_es.ts
@@ -191,7 +191,7 @@ No obstante, el modo hardcore también impide el uso de los estados de guardado,
     </message>
     <message>
         <source>Show Leaderboard Notifications</source>
-        <translation>Mostrar notificaciones de tabla de posiciones</translation>
+        <translation>Mostrar notificaciones de tabla de clasificación</translation>
     </message>
     <message>
         <source>Unchecked</source>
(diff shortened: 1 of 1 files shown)


Pesa:
You should run `lupdate` again with `-locations none` after you've made your changes.

src/duckstation-qt/translations/duckstation-qt_es.ts
```diff
@@ -601,6 +601,34 @@ Error: {}</translation>
         <comment>Pause Menu</comment>
         <translation>Intentos activos de entrar en tabla de clasificación</translation>
     </message>
+    <message>
+        <source>Game was beaten {0}, and completed {1}.</source>
+        <translation>El juego fue superado {0} y completado {1}.</translation>
+    </message>
+    <message>
+        <source>Subset was beaten {0}, and completed {1}.</source>
+        <translation>El subconjunto fue superado {0} y completado {1}.</translation>
```
Pesa: The [official RA translation](https://github.com/RetroAchievements/RAWeb/blob/04b2d0503bc014765ef31a2cb751efaa6b398add/lang/es_ES.json) doesn't translate the word "subset", so you should probably keep "subset" instead of "subconjunto" (here and everywhere else).

PeterDelta: I don't understand why they choose not to translate words that already exist in the target language. For example, it's fine that "frames" or "sprite" aren't translated because there isn't a word in Spanish that adequately describes them, but "subset" does. And in this context, the appropriate word is: "conjunto/grupo".

PeterDelta:
It's ready to merge

PR #3789, opened 2026-08-23
Qt: Group toolbar-related menu items into a submenu

Pesa:
These actions don't need to be used frequently, so let's move them to a submenu to reduce clutter in the View menu, which is already pretty long.


src/duckstation-qt/mainwindow.cpp
```diff
@@ -85,7 +85,7 @@ static constexpr std::pair<const char*, QAction * Ui::MainWindow::*> s_toolbar_a
   {"Reset", &Ui::MainWindow::actionRestartGame},
   {"Pause", &Ui::MainWindow::actionPause},
   {"ChangeDisc", &Ui::MainWindow::actionChangeDisc},
-  {"Cheats", &Ui::MainWindow::actionCheatsToolbar},
+  {"Cheats", &Ui::MainWindow::actionToolbarCheats},
```
Pesa, line 88: forgot to mention... I renamed this for consistency with other toolbar-only actions (they all start with `actionToolbar...`)

